PR #8, opened 2016-08-24
Add tf.split operation

karenl7:
see https://www.tensorflow.org/versions/r0.10/api_docs/python/array_ops.html#split


malmaud:
Thanks!

PR #13, opened 2016-09-03
Allow train.Saver to accommodate namespaced variables

karenl7:


malmaud:
Can you give me some context on what this is trying to fix?

karenl7:
When I was restoring a saved session, an error came up because for my dictionary of variables (var_name, var_value) the var_value was also another dictionary.

The suggested change generalises the restore function to account for cases like the above.
https://github.com/malmaud/TensorFlow.jl/blob/b19cc2b8cba9a162a8eb6f3e78debc4e7b5f1958/src/train.jl#L168

malmaud:
Was the checkpoint you were restoring from created by calling the `save(::Saver, ...)` method? That shouldn't save a dictionary-valued value.

karenl7:
To be specific, this is the behavior that I see:

```julia
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.0-rc3+0 (2016-08-22 23:43 UTC)
 _/ |\__'_|_|_|\__'_|  |  
|__/                   |  x86_64-linux-gnu

julia> using TensorFlow

julia> let sess = Session(Graph())
           Z = Variable(ones(Float32, 2, 2), name = "Z_var")
           X = variable_scope("namespace") do
               get_variable("X_var", [2, 2], Float32)
           end
           saver = train.Saver()
           run(sess, initialize_all_variables())

           save_path = "tf_save/model.txt"
           train.save(saver, sess, save_path)
           info("Checkpoint files saved in $save_path")
       end

julia> using JLD

julia> load("tf_save/model.txt")

Dict{String,Any} with 2 entries:
  "namespace" => Dict{String,Any}(Pair{String,Any}("X_var",Float32[-0.00945762 -0.00221038; -0.0231152 -0.00255698]))
  "Z_var" => Float32[1.0 1.0; 1.0 1.0]
```

malmaud:
Oh, I see - when JLD is told to save a variable called something like "x/y", it creates a new HDF5 group called "x" with "y" in it. 

Anyways, thanks for this fix!

PR #32, opened 2016-09-21
Pull request/7e6f16f3

baggepinnen:
This pull request adds
- Shape inference for `gather`
- Documentation embryos for all functions in `transformations.jl`

Modifies:
- Syntax for `register_shape` such that `do`-block syntax can be used
http://docs.julialang.org/en/release-0.5/manual/functions/#do-block-syntax-for-function-arguments


malmaud:
Thanks! This is really useful.

PR #35, opened 2016-09-28
Add fully_connected layer creation function

baggepinnen:
Add helper function to create fully connected layer


malmaud:
I'm going to close because these higher-level abstractions will live in a new package that uses tensorflow. Hope you can contribute to that, @baggepinnen.

PR #37, opened 2016-09-28
Add autoencoder example implementation

baggepinnen:


baggepinnen:
Btw, this pull request requires #36 to work

kshyatt:
@baggepinnen would this example be suitable to run as a test, the same way we make sure the logistic regression works?

baggepinnen:
Hello @kshyatt , I will have a look at it when I'm back behind my desk in a few days. Nice to see all the activity going on here the last couple of days :)

PR #42, opened 2016-09-30
Juno Display

MikeInnes:
Just a suggestion; I can tweak this as needed to show the right information in the right places.

screenshot 2016-09-30 12 24 38

I also added the `deps` folder to gitignore wholesale because it was showing me download files that didn't need committing – but let me know if that's mistaken.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 deps
+!deps/build.jl
+!deps/build_pytensorflow.jl
 src/scratch.jl
 tfdocs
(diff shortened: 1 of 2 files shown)


malmaud:
Thanks, this is really great! 
I would like to keep deps/build.jl and deps/build_pytensorflow.jl under version control, though.

MikeInnes:
Yeah that does seem sensible doesn't it :) Thanks!

PR #47, opened 2016-10-05
Add troubleshooting

MikeInnes:
This may save you some time if others have the same issue as me.


malmaud:
Thanks, that is very useful

PR #48, opened 2016-10-05
Fix conversion error

MikeInnes:
Prior to this I was getting `no method convert(TensorShape, ::Array{Int32})` coming from line 83, so I guess `shape_inferrer` isn't guaranteed to return a `TensorShape`.

After this I'm still seeing an `invalid Array dimensions` error coming from `get_attr` with shape inference – I'll push a fix if I have one.

### DIFF
--- a/src/shape_inference.jl
+++ b/src/shape_inference.jl
@@ -214,7 +214,7 @@ register_shape("Reshape") do op
     if isnull(maybe)
         return [TensorShape(nothing)]
     else
-        return [get(maybe)]
+        return [TensorShape(get(maybe))]
     end
 end
 


malmaud:
`shape_inferrer` is actually supposed to return a `TensorShape`, so there is a mistake in one of the shape inference functions (which doesn't surprise me - we need more testing for them). Probably better to find that mistake and fix it then sweep it under the rug by converting on line 79.

MikeInnes:
Makes sense. I'll stick with this for now personally so I can get something functional, then perhaps try to dig into the root cause more.

malmaud:
Sure

MikeInnes:
The issue was in the `Reshape` shape inference function. Pushed a better fix.

malmaud:
Great

PR #52, opened 2016-10-06
Wrap way more nn functions

kshyatt:
Hopefully I did **some** of this correctly...

### DIFF
--- a/src/ops/nn.jl
+++ b/src/ops/nn.jl
@@ -204,10 +204,23 @@ end
 @not_implemented function all_candidate_sampler()
 end
 
-@not_implemented function atrous_conv2d()
+@not_implemented function atrous_conv2d(value, filters, rate, padding; name="")
+    desc = NodeDescription("AtrousConv2D", get_name(name))
+    add_input(desc, Tensor(value))
+    add_input(desc, Tensor(filter))
+    desc["padding"] = padding
+    desc["rate"]    = rate
+    Tensor(Operation(desc), 1)
 end
 
-@not_implemented function avg_pool()
+function avg_pool(value, ksize, strides, padding; data_format="NHWC", name="")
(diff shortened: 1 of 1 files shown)


src/ops/nn.jl
```diff
@@ -204,10 +204,23 @@ end
 @not_implemented function all_candidate_sampler()
 end
 
-@not_implemented function atrous_conv2d()
+@not_implemented function atrous_conv2d(value, filters, rate, padding; name="")
```
malmaud: Remove the `@not_implemented` macro now that it's implemented :)

src/ops/nn.jl
```diff
@@ -204,10 +204,23 @@ end
...
+    desc = NodeDescription("AtrousConv2D", get_name(name))
```
malmaud: In the new style, this should be something like

```julia
local desc
with_op_name(name) do
    desc = NodeDescription("AtrousConv2D")
end
```

instead of using the two-argument `NodeDescription` constructor.

src/ops/nn.jl
malmaud: In the new style, a default name should here in the function definition instead of a blank string.

src/ops/nn.jl
```diff
@@ -204,10 +204,23 @@ end
...
+    add_input(desc, Tensor(value))
+    add_input(desc, Tensor(filter))
+    desc["padding"] = padding
+    desc["rate"]    = rate
+    Tensor(Operation(desc), 1)
 end
 
-@not_implemented function avg_pool()
+function avg_pool(value, ksize, strides, padding; data_format="NHWC", name="")
```
malmaud: Same comment about the `with_op_name` apply to all these new functions

PR #60, opened 2016-10-07
Add tests for clipping methods

kshyatt:


test/clipping.jl
```diff
@@ -0,0 +1,19 @@
+using TensorFlow
+using Base.Test
+
+sess = TensorFlow.Session()
```
malmaud: It might be good practice when writing tests to create new sessions with 
`sess = TensorFlow.Session(TensorFlow.Graph())`. That will ensure the tests don't share any part of their computation graph.

PR #61, opened 2016-10-07
Clean up comparison file and add tests

kshyatt:


test/comp.jl
```diff
@@ -0,0 +1,20 @@
+using TensorFlow
+using Base.Test
+
+sess = TensorFlow.Session()
```
malmaud: Let's change this to
`sess = TensorFlow.Session(TensorFlow.Graph())`

PR #64, opened 2016-10-07
Fix Conv2D shape inference

MikeInnes:
This is still in progress as there are further errors, but it at least fixes the first.


kshyatt:
Any way you could add a test for this?

MikeInnes:
Can do – once it actually works I'll just call `get_shape` in a test.

MikeInnes:
The further errors were PEBCAK but I added a couple assertions to make that less likely. Test is applied also

PR #74, opened 2016-10-09
Wrap segmented reductions

kshyatt:


src/ops/math.jl
```diff
@@ -252,3 +252,17 @@ for reduction in [:sum, :prod, :min, :max, :all, :any, :mean]
         end
     end
 end
+
+for reduction in [:sum, :prod, :min, :max, :mean]
+    func_name = ucfirst(string(reduction))
+    @eval function $(Symbol("segment_", reduction))(n::AbstractTensor, segment_indices; name="Segment"*$func_name)
+        segment_indices = [Int32(idx-1) for idx in segment_indices]
```
malmaud: The segment indices might be a tensor instead of a constant, right? In which case this won't work.

kshyatt: Should I convert to a `Tensor`, then `cast` and `.- 1`?

malmaud: Yep

PR #77, opened 2016-10-10
Fix shape calculation shape calculation

MikeInnes:
shape calculation


MikeInnes:
This now includes an unrelated commit allowing `reshape(x, dims)` to work when `dims` is a tensor. Can split that out if it's better for review though.

PR #97, opened 2016-10-30
replace `@test` with `Base.Test.@test` in first example

oxinabox:
That way, the example can be copy pasted, and ran,
without any issues.


malmaud:
Thanks!

Issue #113, opened 2017-01-04
shape_inference of split


karenl7:
I suspect a possible typo. `spit_dim` is not defined.
https://github.com/malmaud/TensorFlow.jl/blob/master/src/shape_inference.jl#L452

malmaud:
Fixed in https://github.com/schmrlng/TensorFlow.jl/commit/12a1f1a2a3ed17bdebfcda7f2c1bbd168e01c03a

PR #114, opened 2017-01-09
Fix shape inference for split #113

schmrlng:
The necessary `+1` on the dimension index seems to be in line with many of the other functions in shape_inference.jl.


malmaud:
Thanks!

PR #125, opened 2017-02-01
Fix Unpack, and refactor to use `get_shape(x, dim)`

oxinabox:
unpack was broken for three reasons;
1. if `num` was not provided the path to calculate it used `size` rather than `get_shape`
2. `axis` was 0-based not 1 based (this was also wrong in `pack` so that was fixed too)
3. The way defaulting `num`/`axis` works is not part of the tensor flow operation AFAICT, and need to be done by the wrapper.

In solving 1, i defined a `get_shape(x, dim)` that corresponds to `size(x, dim)`.
And throws exceptions if the shape is unknow or the shape for that axis is unknown.
I then went through and replaces the handful of times this pattern was used.
Arguable the new error message is less clear. But also arguably it is more specific.
If you do not like, you could cherry pick just the first commit.

This is a breaking change, because of the `axis` argument becoming 1 based.

### DIFF
--- a/src/ops/nn.jl
+++ b/src/ops/nn.jl
@@ -117,14 +117,7 @@ function rnn(cell, inputs; initial_state=nothing, dtype=nothing, sequence_length
         if dtype === nothing
             error("dtype must be set if initial_state is not provided")
         end
-        shape = get_shape(inputs[1])
-        if shape.rank_unknown
-            error("Shape of input is unknown")
-        end
-        if isnull(shape.dims[1])
-            error("Batch size of input is unknown")
-        end
-        batch_size = get(shape.dims[1])
+        batch_size = get_shape(first(inputs), 1)
         initial_state = zero_state(cell, batch_size, dtype)
     end
     outputs = Tensor[]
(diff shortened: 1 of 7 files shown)


oxinabox:
It it right that `unpack` should cause all shape information of the tensors to be lost?
It seems wrong to me, the shape should be statically infer-able -- an example of it is in the docs.

staticfloat:
It seems to me that it might be useful to have `get_shape(x, dim)` return `1` if the rank of `x` is less than `dim`. This allows more generic thinking about shapes of things, so that, for instance, we don't run into problems like I ran into when trying to multiply a matrix`W` and a vector `v` as in `W*v`, and finding out that [`MatMul` always expects a matrix on the right hand side](https://github.com/malmaud/TensorFlow.jl/blob/a6b656176de3978dc9bee572263a33f73338ec47/src/shape_inference.jl#L243).

oxinabox:
>It seems to me that it might be useful to have get_shape(x, dim) return 1 if the rank of x is less than dim

That is a different behavior.
I would call that `get_rank` as the static version of `rank`, matching the relationship between `get_shape` and `shape`.
And would write `get_rank(x)>=dim`

`get_shape(x, dim)` is useful in and of itself, I think it is used in this PR in about 5 places, to replace repetitive code.

malmaud:
Thanks, this is really good!

PR #129, opened 2017-02-08
Give `squeeze` a default dimension, similar to the python API

staticfloat:
This allows one to use `squeeze(z)` to automagically get rid of those pesky `1` dimensions.


malmaud:
Thanks!

PR #132, opened 2017-02-11
Improve Shape Inference for things that take lists of Tensors

oxinabox:
This PR fixes shape inference for `concat` and `pack`

 - `concat` would fail at inference if a vector with some dimensions unknown was first
 - `pack` wouldn't do type inference at all

It also leaves  shape inference for `add_n` wrong, but in a different way.
It is still not broadcasting, but now it gives errors instead of being wrong in broadcast scenarios.
It really wouldn't be too hard to make it correct.
It is just the same generalisation for unification to make a multiargument broadcast.
Which I could do in this PR, or I could do it later seperately.

Also it makes Concat 1 based. Because that was killing me every time I used it.
I guess a deprecation strategy could be employed for that.
Like deprecating concat infavor of cat (which was alway correct).
I put that in it's own commit so it could be cherry picked out. (so long as the testing cat commit is kept)
But I am not sure having it around at all is a good idea -- it just allows for confusing code.

### DIFF
--- a/src/ops/transformations.jl
+++ b/src/ops/transformations.jl
@@ -155,20 +155,25 @@ end
 """
 concat(dim, values; name="")
 
+Concatenates the list of tensors `values` along dimension `dim` (1-based).  If
+`values[i].shape = [D1, D2, ... Daxis(i), ...Dn]`, the concatenated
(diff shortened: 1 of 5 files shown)


malmaud:
Thanks! This is good stuff.

PR #141, opened 2017-02-16
=Shape inference and 1-based indexing for ExpandDims

oxinabox:
ExpandDims now has `get_shape` working.
(Before it just always returned `[Tensor(nothing)]`)

And it has now has 1 based indexing.


malmaud:
This has to rebased, but otherwise looks good.

oxinabox:
Right rebased.
Is there a better way to do parallel branches that doesn't cause this to happen?

malmaud:
Not that I know of, unfortunately.

PR #144, opened 2017-02-18
gather_nd

oxinabox:
This PR adds support of [GatherNd](https://www.tensorflow.org/versions/r0.10/api_docs/python/array_ops/slicing_and_joining#gather_nd)

Using it for slicing is a really weird to me most of the time. (though I think I have a really handy use for it for what I am actually doing right now).
Slice is more sane to me a lot of the time.

But what this does give us is indexing.
So now `X[2,3]` will work. (Cool but maybe not super useful)
No slicing in `get_index` yet; I have some thoughts toward it.

### DIFF
--- a/src/ops/transformations.jl
+++ b/src/ops/transformations.jl
@@ -383,7 +383,8 @@ end
 
 """
 Base.size(n::AbstractTensor; name="")
-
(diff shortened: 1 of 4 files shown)


malmaud:
Thanks!

PR #145, opened 2017-02-21
Be smarter about collections

staticfloat:
This allows you to pass in new collection names into things like summaries, then get them back out with `merge_all_summaries()`.


malmaud:
Thanks!

PR #149, opened 2017-02-23
Fix Gather so its grad can be calculated

oxinabox:
I am not 100% sure why, but it seems that when GatherNd is differencated to ScatterNd, the shape comes back as Int32, and the indicies as the type they were on GatherNd.

[ScatterNd](https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/scatter-nd) requires that it's shape input and it's gather input have same type.

This fixes that


malmaud:
Great, thanks.

Issue #147, opened 2017-02-22
Concat is broken on master


oxinabox:
<s>

MWE:

```julia
sess = Session(Graph())
x = constant(rand(3))
run(sess, cast(x, Int32))
```

Output:

```
Tensorflow error: Status: NodeDef mentions attr 'T' not in Op<name=Cast; signature=x:SrcT -> y:DstT; attr=SrcT:type; attr=DstT:type>; NodeDef: Cast = Cast[DstT=DT_INT32, SrcT=DT_DOUBLE, T=DT_INT32, _class=[]](Const)


 in check_status(::TensorFlow.Status) at /home/ubuntu/.julia/v0.5/TensorFlow/src/core.jl:231
 in TensorFlow.Operation(::TensorFlow.NodeDescription) at /home/ubuntu/.julia/v0.5/TensorFlow/src/core.jl:625
 in #cast#378(::Void, ::Function, ::TensorFlow.Tensor, ::Type{T}) at /home/ubuntu/.julia/v0.5/TensorFlow/src/ops/transformations.jl:37
 in cast(::TensorFlow.Tensor, ::Type{T}) at /home/ubuntu/.julia/v0.5/TensorFlow/src/ops/transformations.jl:29
```
</s>

oxinabox:
Hmm, I am having trouble reproducing.
May have been a result of not re-running `build.jl`

I can produce a similar error from other code, but not this MWE anymore.

malmaud:
Maybe caused by https://github.com/malmaud/TensorFlow.jl/blob/84403b5be10cab777d3c25c645f65b964aa3ab8b/src/core.jl#L203? Hard to be sure without a MWE, but I'll alter that logic at least.

malmaud:
I made the change - let me know if you can still find a reproducible example.

oxinabox:
I can still reproduce it.
I am trying to make a MWE.
It is proving hard.

oxinabox:
Here we go.
Subtracting 1 is broken.

```julia
using TensorFlow
sess2 = Session(Graph())
overshot_targets = constant(rand(1:9, 5)+1)
w= get_variable("w", (5,10), Float64)
targets = overshot_targets - 1

costs = nn.sparse_softmax_cross_entropy_with_logits(w, targets)
cost = reduce_mean(-costs) #cross entropy
optimizer = train.minimize(train.AdamOptimizer(0.1), cost)
run(sess2, initialize_all_variables())

[...]

PR #150, opened 2017-02-23
Breaking Testcase for #147

oxinabox:
Here is the breaking testcase for #147 ,
together with the debug code.
If this should prove to be a better starting point


malmaud:
Thanks for drilling down. I'll debug.

oxinabox:
Seems fixed

PR #151, opened 2017-02-23
Not using globals to get results from pyproc (minor refactor)

oxinabox:
There is no need to create global variables in Main here.

Ideally there would be away to do this without using `eval` to change scope at all, because that not nice.
But right now i can't see how.


malmaud:
I agree; I hate having to use `eval` there. 

This is a welcome PR. I can't remember the original reason I did it the other way.

PR #155, opened 2017-02-25
Mention building from source to get around CUDA version

oxinabox:
I now have TensorFlow.jl working with CUDA 7.5.
Yay


malmaud:
Looks good. Just in case you aren't aware, it's perfectly possible to have multiple CUDA versions installed on the same system at the same time - I've been using TensorFlow with CUDA 8.0 and caffe with 7.5.

PR #157, opened 2017-02-26
Use libtensorflow.so instead of libtensorflow_c.so.

asimshankar:
The two are identical but the former is recommended by the
TensorFlow maintainers and is created in binary releases of
the TensorFlow C library.

The existence of both targets is an unfortunate outcome of tensorflow/tensorflow@783c52e and tensorflow/tensorflow#3830 happening within days of each other.

Note that the TensorFlow C library is now included in binary releases of TensorFlow. The releases are available in URLs of the form:   `https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-TYPE-OS-ARCH-VERSION.tar.gz`. For example:

- CPU-only, Linux, x86_64, 1.0.0: https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-linux-x86_64-1.0.0.tar.gz
- GPU-enabled, Linux, x86_64, 1.0.0:
https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-gpu-linux-x86_64-1.0.0.tar.gz
- CPU-only, OS X, x86_64, 1.0.0:
https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-darwin-x86_64-1.0.0.tar.gz
- GPU-enabled, OS X, x86_64, 1.0.0:
https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-gpu-darwin-x86_64-1.0.0.tar.gz

Perhaps you want to consider using these to speed up getting started with TensorFlow.jl where possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,8 +113,8 @@ sudo pip3 install --upgrade $TF_BINARY_URL
 If you want to build your own version of the TensorFlow binary library instead of relying on the one that is installed with
 `Pkg.build("TensorFlow")`, follow the instructions from https://www.tensorflow.org/install/install_sources, except:
 
-* In the section "Build the pip package", instead run `bazel build --config=opt //tensorflow:libtensorflow_c.so`.
-* Then copy the file "bazel-bin/tensorflow/libtensorflow_c.so" to the "deps/usr/bin" directory in the TensorFlow.jl package.
-* On OS X, rename the file to `libtensorflow_c.dylib`.
+* In the section "Build the pip package", instead run `bazel build --config=opt //tensorflow:libtensorflow.so`.
+* Then copy the file "bazel-bin/tensorflow/libtensorflow.so" to the "deps/usr/bin" directory in the TensorFlow.jl package.
+* On OS X, rename the file to `libtensorflow.dylib`.
 
 A convenience script is included to use Docker to easily build the library. Just install docker and run `julia build_libtensorflow.so` from the "deps" directory of the TensorFlow.jl package.
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -72,6 +72,8 @@ function try_unzip()
     end
 end
 
+const cur_version = "1.0.0-rc1"
+
 @static if is_apple()
     r = Requests.get("https://storage.googleapis.com/malmaud-stuff/tensorflow_mac_$cur_version.zip")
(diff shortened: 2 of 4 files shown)


oxinabox:
The download URLs in `/deps/build.jl` need to be updated.
This is why travic-CI is failing.
Because it downloads a zip containing `libtensorflow_c.so`,
and tries to move that into possition.
But you have updated the move command to try and move `libtensorflow.so`
(and same for other mac.)

You could also update the download URL to the official ones you provide, which contain little more than the the .so file, and a header file. but inside a folder.
This seems like a good idea to me.
Saves on @malmaud manually uploaded the file to his storage.

malmaud:
This is great - I didn't know the official binaries now included the C library! 

As @oxinabox says, do you want to update the build.jl to use the official Google binaries? Or I'm happy to do it.

asimshankar:
Thanks @oxinabox and @malmaud. I've updated `deps/build.jl` to use these URLs.

BTW, why is there a dependency on the Python libraries? (`deps/build_pytensorflow.jl`). Could that be removed? I notice that it is still installing 0.12 and not 1.0.

malmaud:
TensorFlow.jl farms out to Python for two things at the moment: gradient calculations and writing out summaries for use with TensorBoard. I'd love to remove that dependency, but those things aren't part of the C API at the moment. 

If you don't actually use those functions, then you don't need the Python libraries since they are lazily loaded.

asimshankar:
Thanks for the info. Could you help me with the Travis failure for this PR?

malmaud:
Ah ya, you're right. I guess all we can is put this on hold until the next upstream release.

malmaud:
How about keeping the parts of the PR that change tensorflow_c.so to tensorflow.so, but removing the part that uses TensorFlow official binaries?

asimshankar:
Good idea, done.

I've left the code to use the release version in the file, but commented out. Happy to remove it if you think that is best.

Please take another look.

asimshankar:
@malmaud : ping?
(FYI: The `libtensorflow_c.so` target is no more: https://github.com/tensorflow/tensorflow/commit/89be13427b247d4f10b0b22d0c92ca6d3600f494)

malmaud:
Looks good. Thanks for submitting this!

PR #158, opened 2017-03-01
Fix shape inference for Where/Find 

oxinabox:
Where/Find can alway infer the second dimenstions shape at compile time.
But never the first.
So this makes that happen.

Very concerningly when I run `julia test/shape_inference.jl`
all tests pass.
Where as if I run `julia test/runtests.jl`
I get:
```
shape_inference: Test Failed
  Expression: get_shape(find(placeholder(Bool; shape=[10,20,30]))) == TensorShape([-1,3])
   Evaluated: TensorShape[?, 10] == TensorShape[?, 3]
 in record(::Base.Test.DefaultTestSet, ::Base.Test.Fail) at ./test.jl:428
```

### DIFF
--- a/src/shape_inference.jl
+++ b/src/shape_inference.jl
@@ -657,7 +657,8 @@ register_shape("Where") do op
     if shape.rank_unknown
         [TensorShape(nothing)]
     else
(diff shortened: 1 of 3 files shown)


malmaud:
Thanks, I fixed up the bug with runtests.jl and manually merged.

PR #160, opened 2017-03-02
ScatterNd

oxinabox:
This PR implements the ScatterNd op.
Which is is to `set_index` as GatheNd is to `getindex`

Nice thing about this is it lets me implement masking on the leading dimensions as something like

```julia
function direct_mask(V, mask)
    inds = find(mask)
    flat_masked = gather_nd(V, inds)
    scatter_nd(inds, flat_masked, shape(V))
end
```

I've not run that exact code, I have stuff that operates on flatmax -- like my loss function.
and I can use `get_shape`.

But ScatterNd seems useful


malmaud:
Awesome

PR #164, opened 2017-03-06
add two ops

baggepinnen:
had to make a new branch with the batch_matmul
added `squared_difference`


oxinabox:
Can these have tests?
and entries in `src/shape_inference.jl`, and (tests for those).

malmaud:
Thanks! On the newer versions of upstream TensorFlow, `batch_matmul` has been rolled into `matmul` so I put in a deprecation warning.

PR #165, opened 2017-03-08
.!= was broken on julia 0.5. 

oxinabox:
https://github.com/malmaud/TensorFlow.jl/commit/bd28eb2e66cf2f0cbff308a0d6d4165e1fa81cb7 broke it I think.

I've added tests for each infix op, so as to be sure they are all imported still.
This might break 0.6
I'll let TravisCI tell me.


malmaud:
Seems like the CI failure is a problem with Homebrew building HDF5, cf https://github.com/JuliaPackaging/Homebrew.jl/issues/180

malmaud:
Looks like this is fine; CI is just a problem with Homebrew

Issue #102, opened 2016-11-06
Define `size` for things with `get_shape`


oxinabox:
`get_shape` seems to be TensorFlow for `Base.size`.
But with the ability for it to be unknown etc.

Eg
```
get_shape(Ts) 
Out[93]:
TensorShape[407, 16]
In [96]:

get_shape(X_ids)
Out[96]:
TensorShape[unknown]
```

Does it makes sense to define a overload for `Base.size`, that returns a tuple of `Ints`, or throws an exception if the size is unknown?

malmaud:
I'm reluctant to do that since `size(::Tensor)` already has a meaning in Julia - the Julia 'size' of any non-container object is `()` and it could get confusing to alter that meaning for tensors. But I am open to counter-arguments.

IainNZ:
+1 for not conflating the two, the details are sufficiently different that I think it could only be confusing.

malmaud:
OK, I'll close for now.

PR #166, opened 2017-03-11
Implemented Slicing Indexing (inc end)

oxinabox:
This PR beings support for slice indexing for `getindex`
Eg `X[1:3,:, 8:end]`

It does not bring the mixed-mode indexing.
though I am starting to have more and more notions of how that would be done.

This is a breaking change as it redefines `Base.length` to do the `Size` op, and `Base.size` to do the `shape` op.

This is, effectively also the strong counter arguement to not-doing https://github.com/malmaud/TensorFlow.jl/issues/102
though now with dynamic `shape`, rather than static `get_shape`

It is required as like `endof(X)` is in linear indexing,
`size(X,dim)` is required in cartesean indexing 

Technically could define `size(X,dim)` without defining `size(X)` to match, but i feel that is more uninituitive, than any other option.
This is requires because it is hardcoded to replace `end` during parsing (lowering?)
https://github.com/JuliaLang/julia/blob/75c2e72c505be06e1f80498d3746a62a4f239237/src/julia-syntax.scm#L104

CrossRef, what is going on for indexing upstream / in python.
https://github.com/tensorflow/tensorflow/issues/206

Do review this carefully.
I expect it needs some style changes still.
and also it does change things in a few more places than may be expected so might have broken something.

I think a refactor of the usage of  `Int32` vs `cast` vs `convert_number` vs `convert(Int32,..)`
for the  whole code base is in oder

### DIFF
--- a/src/ops/transformations.jl
+++ b/src/ops/transformations.jl
@@ -578,7 +578,7 @@ function Base.getindex(params::AbstractTensor, indices...)
     if eltype.(inds) ⊆ (Int32, Int64)
         gather_nd(params, inds)
     else
-        error("julia style indexing is not currently supported for indicies $inferred")
+        error("julia style indexing is not currently supported for indicies $indices")
     end
 end
 
--- a/src/ops/transformations.jl
+++ b/src/ops/transformations.jl
@@ -767,7 +767,7 @@ mask = [True, False, True]
 boolean_mask(tensor, mask) ==> [[1, 2], [5, 6]]
 ```
 """
-@op function boolean_mask(tensor, mask; name=nothing)
+@op function boolean_mask(tensor, mask::AbstractTensor; name=nothing)
     local result
     with_op_name(name, "BooleanMask") do
         indices = find(mask)  # TODO generalize to more dimensions
@@ -776,6 +776,17 @@ boolean_mask(tensor, mask) ==> [[1, 2], [5, 6]]
     end
     result
 end
+
+@op function boolean_mask(tensor, mask::AbstractArray; name=nothing)
+    local result
+    with_op_name(name, "BooleanMask") do
+        indices = find(mask)  # TODO generalize to more dimensions
+        result = tensor[indices]
+    end
+    result
+end
+
+
 
 """
 `transpose(n::AbstractTensor, perm=nothing)`
--- a/src/ops/transformations.jl
+++ b/src/ops/transformations.jl
@@ -522,14 +522,14 @@ end
 
 # Colon promotion rules -- mostly this will make Ints into constants
 immutable TensorRange
-    start::Tensor
-    stop::Tensor
+	start::Tensor{Int32}
+	stop::Tensor{Int32}
 end
 Base.first(tr::TensorRange)=tr.start
 Base.last(tr::TensorRange)=tr.stop
 
 Base.colon(x,y::Tensor) = colon(Tensor(x), y)
-Base.colon(x::AbstractTensor, y) = colon(Tensor(x), Tensor(y)) #HACK brake symetry by using AbstractTensor for 1 and Tensor for the other
+Base.colon(x::Tensor, y) = colon(x, Tensor(y))
 Base.colon(x::Tensor,y::Tensor) = TensorRange(x,y)
 
 #For x[[1,2,3]] etc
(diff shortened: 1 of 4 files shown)


oxinabox:
Oh dear,  I forgot to force push.
I'll re-rebase this in a bit

oxinabox:
One interesting thought that occurs to me, 
is that `size` could first attempt to do static `get_shape`.
And then if that had any unknowns or had unknown rank,
it could then fall back on the dynamic `shape`.
This would make it type unstable; but I'm not sure we care.

In the case of `size(X,dim)` it could only need to fall back the `shape` if the `dim` axis was unknown (or the rank) was unknown.

What do you think?

BTW: I think the CI failure has nothing to do with this PR ?

oxinabox:
Updated for #176 
That has made the code a lot cleaner.

src/ops/transformations.jl
```diff
@@ -518,17 +543,50 @@ function Base.getindex(params::AbstractTensor, indices)
...
+function Base.getindex(params::AbstractTensor, indices::Vararg{Union{TensorRange, UnitRange, Colon}})
+    # This function is all about slices
+    # NOTE: slice is still 0 based for begins
+
+    # TODO: Assign a name prefix to all the tensors made as art of this section, including constants
+    begins = Tensor{Int32}[]
+    sizes = Tensor{Int32}[]
+
+    function proc_ind!(ind::Colon)
+        push!(begins, 0)
```
malmaud: I updated slice to be 1-based, so this should adjust now too.

src/ops/transformations.jl
```diff
@@ -518,17 +543,50 @@ function Base.getindex(params::AbstractTensor, indices)
...
+        push!(sizes, -1) # Slice mark for go to end
+    end
+    function proc_ind!(ind::Union{UnitRange, TensorRange})
+        #NOTE: end has now been replace with `endof(X)` or `size(X,d)` giving the actual size
+        begin_ =  first(ind) - 1
```
malmaud: And here

oxinabox:
Done with fixing the 1 based vs 0 based indexing.
I also moved all this to its own file, because it is a bit distinct from the other transformations, and that file was getting rather long.
(I'm sure you can copy paste back if you do not like).


What are your thoughts on making 
on defining `size` as something like:

``julia
@op function Base.size(n::AbstractTensor; name=nothing) = shape(n; name=name)
     static_shape = get_shape(n)
     if static_shape.rank_unknown || any(isnull.(shape.dims)
             shape(n)
     else
            static_shape
     end
end


@op Base.size(n::AbstractTensor, i; name=nothing)
     static_shape = get_shape(n)
     if static_shape.rank_unknown || isnull(shape.dims[i])
             shape(n)[i]
     else
            static_shape[i]
     end
end
```

**Comment by malmaud** [comment]
Mmm, so it either returns a literal Int tensor if the shape is inferable, and otherwise a Tensor representing the shape? Why don't we try it out for now and see if it causes any difficulties that need it to be reverted down the line?

**Comment by oxinabox** [comment]
ok then; I'll make that change some time tomorrow (It is very late now in Australia).
Key downside I can see from the get-go is that it is type-unstable.

**Comment by oxinabox** [comment]
I'm less sure now doing mixing `get_shape` with `shape` is a good idea.
When `get_shape` works, the answer will be in `Int32`/ `Vector{Int32}`,
when `shape` is required the answer will be in `Tensor{Int32}`.
Julia `Int32` has different rounding behaviour from a `Tensor{Int32}`.
Namely that `3/2 ==1.5`. Vs `constant(3)/constant(2) == constant(1)`

The work around would be to wrap the results of `get_shape` into `constant`s.
And so have any math do to them inserted into the Graph.
But at that point, is this really any different from just adding a `shape` call to the graph?
I suspect performance-wise it is not.
And it may be worse if it means that we bipass any pre-execution graph optimisation stages that might target `shape` calls (idk if there are any optimistation passes at all though. I know _theano_ has them.)

**Comment by oxinabox** [comment]
For reasons I no-longer properly recall,
today I made it possible to do things like:

```julia
@test w_jl[2:4, 3, :] ==  run(sess, w[2:4, 3, :])
@test w_jl[:, 3, :] ==  run(sess, w[:, 3, :])
@test w_jl[:, end, :] ==  run(sess, w[:, end, :])
@test w_jl[1:1, :, :] ==  run(sess, w[1:1, :, :])
```
Which is pretty swish.

This means we support mixed mode *indexing*, and *slicing*.
We don't support mixing in masks for some dims. (Or masks beyond 1D)
Supporting mixing in masks, and mixing in index-lists (for lack of a better term) is all that is left for get-index. I still think #183 would be useful there.

I fixed  up some mask issues in  4ccd0f2 BTW, while I was there. Tests are added in  ce07eec

malmaud:
Feel free to merge when ready.

oxinabox:
Since it currently  has no conflicts.
And I have been using it for a while (on various versions).
I'll just merge it now then.
Later enhancements can come later.

PR #172, opened 2017-03-15
Correct default dimension for LSTMCell

oxinabox:
I feel like this was meant to default to -1 rather than 1 ie input_dim=-1.
(rather unexpected break in my code from this change)

regression from 
https://github.com/malmaud/TensorFlow.jl/commit/b7e719742d09c5b1aa66bdb46b6fd68724e71645


malmaud:
Yep

Issue #130, opened 2017-02-09
Automatically Name nodes that are assigned to variables


oxinabox:
I thought it would be a lot easier to read debug messages if all the nodes that were assigned to variables could have their name set to their variable name.

I have made a "quick" proof of concept as an IJulia notebook.

https://gist.github.com/oxinabox/15c32bc7c21cb6d8ba507ec5144615e5

If you like the idea, I can finish it off and make a PR.

It is currently a bit insane.

oxinabox:
I have finished up that gist and updated it.
https://gist.github.com/oxinabox/15c32bc7c21cb6d8ba507ec5144615e5

It will not work without https://github.com/malmaud/TensorFlow.jl/pull/131

Right now, this will only work on 0.5.
Making it work on 0.6 requires a (subtly) different reflection method to identify which functions take name keyword arguments.
Making this robust against version changes requires a stable reflection api (https://github.com/JuliaLang/julia/issues/20555).
A work around might be to run the reflection to generate the list of functions to cover, during development. and then serialize (JLD) them into a file, that is checked into the repo.
Which might be cleaner; though it would need updating every time a new function was added etc.

oxinabox:
@malmaud thoughts?
Is this a path worth pursuing?

malmaud:
I need a little time to look at it.

oxinabox:
I see https://github.com/malmaud/TensorFlow.jl/commit/8ce3bbbe0b66397eaa0cb3cb6e8713fda1e96a23

malmaud:
Yep, I've starting playing with the idea. My hope is that long-term `@tf` will do various interesting metaprogramming things, such as transform Julia `while` loops into TensorFlow `while_loop`s. 

A PR to make it as powerful as your gist is in terms of being applicable to a whole nested block would be great.

oxinabox:
Can do.
Do you want me to remove the MacroTools dependency?
It only actually saves a handful of lines, but it does make the code cleaner.

Also what should be done about determining which functions it applies to?
I think including them in a generated file is probably safest, in terms of making it work for all julia versions.
but hand coding the reflection for each julia version is possible.

malmaud:
Including `MacroTools` is fine. 

I'm going to create a new `@op` macro soon that will be prepended to all the TensorFlow.jl functions that define operators, which will automatically give us a database of the functions that a naming macro should apply to.

oxinabox:
[...]

PR #174, opened 2017-03-16
Use Tensor Names better (re #130)

oxinabox:
This is the enhancement to `@tf` discussed in #130
It is largely based on @MikeInnes method.

The extra trick is that we assign a Holy trait to all operations functions (and outer constructors).
To say they are part of TF and should have name kwargs added.


oxinabox:
The second part of this is to make using those names to get hold of the Tensors easier.

> What this code does is allow you to index into a graph (or into a session),
> using the name of a Node.
> And do all the things.
> See: @testset "Graph Node Access By Name"
> 
> This means that only the Graph/session object needs to be passed to a function, in order to interact with the Graph/Session.

src/TensorFlow.jl
```diff
@@ -99,6 +99,8 @@ while_loop,
 get_tensor_by_name,
 as_default,
 @tf,
+withname,
+tf_while,
```
malmaud: I'd rather leave these unexported - ideally, people only indirectly with them via `@tf`.

src/core.jl
```diff
@@ -1230,6 +1230,19 @@ Throws a `NodeNameNotFound` exception if there is no such tensor.
     return Tensor(node, port)
 end
 
+Base.getindex(graph::Graph, name) = get_tensor_by_name(graph, name)
+Base.values(graph::Graph) = get_operations(graph)
+Base.keys(graph::Graph) = (node_name(op) for op in values(graph))
+Base.haskey(graph::Graph, name) = isnull(get_node_by_name(graph, name))
+
+# delegate these operations on the Graph for the session that holds it.
+Base.values(sess::Session) = values(sess.graph)
```
malmaud: I like defining these for `Graph`, but am not sure about `Session` - if we're going to treat `Session`s as dictionaries, maybe the keys should refer to session properties instead of tensors. Maybe let's just comment this out for now.

src/meta.jl
```diff
@@ -1,27 +1,54 @@
...
-            params = Expr(:parameters)
-            insert!(call.args, 2, params)
-        end
-        push!(params.args, Expr(:kw, :name, string(name)))
-    end
+using MacroTools
+
+# Use Holy traits to make if something is a known Op or not
+abstract OpRegistration
+immutable RegisteredOp<:OpRegistration end
```
malmaud: Add space before and after `<:`

src/meta.jl
```diff
@@ -1,27 +1,54 @@
...
+immutable NotRegisteredOp<:OpRegistration end
+
+is_registered_op(::DataType) = NotRegisteredOp() # By default nothing is registered
+
+
+macro op(f)
```
malmaud: I like this new definition, but can we still keep the list of ops like the old version had? It might be useful down the line for something else.

src/meta.jl
```diff
@@ -1,27 +1,54 @@
...
+
+
+# How to insert the  name into functions etc
+# This function takes in a function and its posible name
+# and returns a new function that will call to the orginal
+# with the function name inserted as appropriate
+
+
+withname(::typeof(get_variable), name) = (args...; kwargs...) -> begin
+    if typeof(args[1])<:AbstractString
```
malmaud: Use `isa` here

src/meta.jl
```diff
@@ -1,27 +1,54 @@
...
+        get_variable(args...; kwargs...)
+    else # No name provided
+        get_variable(name, args...; kwargs...)
+    end
+end
+
+withname{F}(f::F, name) = withname(is_registered_op(F), f, name) # Should be at compile time converted to one of the two below
+withname(::NotRegisteredOp, f, name) = (args...; kws...) -> f(args...; kws...)
+withname(::RegisteredOp, f, name) = (args...; kws...) -> begin
+    if !any(first.(kws) .== :name) #name is not already there
```
malmaud: Add space after `#`

src/meta.jl
```diff
@@ -88,24 +115,16 @@ end
...
-    is_assign(arg) = false
-    if ex.head == :block
-        res = Expr(:block)
-        for arg in ex.args
-            if is_assign(arg)
-                push!(res.args, _tf(arg))
+    if ex.head == :while
+        tf_while(ex)
+    else
+        MacroTools.prewalk(ex) do x
```
malmaud: Can you add a comment here clarifying what this `prewalk` is doing?

src/TensorFlow.jl
oxinabox: I'ld rather leave these un-exported too.
I was having namespace issues inside the `@tf` macro.
When the `@tf` macro was run from the `.nn` submodule etc.
I'll give fixing that another shot.

src/meta.jl
oxinabox: We could.
Or we could bring it back when it is needed.
Or we could dynamically generate it using `methods(is_registered_op)`

src/meta.jl
oxinabox: Will do. It is basically doing a recursive find and replace, with the replacement also being searched within. I think. (This is the bit I stole from @MikeInnes )
See also https://github.com/MikeInnes/MacroTools.jl/issues/34

src/core.jl
oxinabox: I see now.
In my own code I use `Session`s when I should in fact just be using the `Graph` at lot of the time, I think.
I basically look at `Session`s as things that hold `Graph`s.
I'll delete it, rather than commenting it out.
Can always dig it out of git history if required.

src/TensorFlow.jl
oxinabox: fixed

src/meta.jl
oxinabox: fixed

src/meta.jl
oxinabox: fixed

src/meta.jl
oxinabox: fixed

src/meta.jl
oxinabox: fixed

src/core.jl
oxinabox: fixed

oxinabox:
just added another commit.
Which resulted from me doing a negative test for DataTypes.
the TL;DR; is that attaching traits to both Function and Outer constructors is hard.
(Because Outer constructors always parse as DataTypes).

PR #176, opened 2017-03-16
Add convert method

malmaud:
Allows for a lot of cleanup


oxinabox:
I was speculating to @deyandyankov
about making all Tensors be `Tensor{eltype, rank}` (with some -1 for rank unknown).
so they could be used with multiple dispatch.

I think it will make fully general julia style indexing code easier to write.

PR #180, opened 2017-03-18
Sequence Length for  RNNs working 

oxinabox:
This makes the Sequence Length parameter actually do something for Dynamic RNNs.
Also the time-major parameter, while I was here.

I can't think of a nicer way to do this right now.
Though it does seem to have excessive code duplication.
I didn't want to bring the `isa(sequence_length, Void)`, inside the `@tensorflow`.
As I don't think lowering / specialisation on input types  work there.
(Or indeed that it will work within closures, on parameters of the parent type, at all)

### DIFF
--- a/src/ops/nn.jl
+++ b/src/ops/nn.jl
@@ -115,7 +115,7 @@ Args:
 * `sequence_length`: Specifies length of each sequence in `inputs`.
 * `scope`: `VariableScope` for the subgraph. Defaults to `RNN`.
 """
-function rnn(cell, inputs; initial_state=nothing, dtype=nothing, sequence_length=nothing, scope="RNN")
+function rnn(cell, inputs::Vector{Tensor}, sequence_length=nothing; initial_state=nothing, dtype=nothing, scope="RNN")
     # TODO use sequence length
     if initial_state === nothing
         if dtype === nothing
@@ -129,9 +129,16 @@ function rnn(cell, inputs; initial_state=nothing, dtype=nothing, sequence_length
     state = initial_state
     for (idx, input) in enumerate(inputs)
         variable_scope(scope; reuse=idx>1) do
-            output, state = cell(input, state)
+            new_output, new_state = cell(data, state, input_dim)
         end
-        push!(outputs, output)
+        if sequence_length!==nothing # This should be removed by the julia lowering process
+            # Only update output and state for rows that are not yet passed their ends
+            have_passed_end = sequence_length .< time_step
+            new_output = select(have_passed_end, output, new_output)
+            new_state = select(have_passed_end, state, new_state)
+        end
+        state = new_state
+        push!(outputs, new_output)
     end
     return outputs, state
 end
--- a/src/ops/nn.jl
+++ b/src/ops/nn.jl
@@ -176,7 +176,7 @@ all elements but all later dimensions may vary.
     num_steps = convert(Tensor{Int64}, tf.shape(inputs)[2])
 
     while_output = @tf while time_step ≤ num_steps
-        slice_start = tf.stack([0, time_step-1, 0])
+        slice_start = tf.stack([1, time_step, 1])
         slice_size = tf.stack([-1, 1, -1])
         data = tf.slice(inputs, slice_start, slice_size)
         data = tf.squeeze(data, [2])
--- a/src/ops/nn.jl
+++ b/src/ops/nn.jl
@@ -175,41 +175,27 @@ all elements but all later dimensions may vary.
     time_step = tf.constant(1)
     num_steps = convert(Tensor{Int64}, tf.shape(inputs)[2])
 
-    while_output = if isa(sequence_length, Void) #This should be removed by the julia lowering process
-        #Don't use seq_length
-        @tf while time_step ≤ num_steps
-            slice_start = tf.stack([0, time_step-1, 0])
-            slice_size = tf.stack([-1, 1, -1])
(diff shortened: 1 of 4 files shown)


oxinabox:
I have been informed that closures do in fact specialise correctly.
So this last commit takes advantage and brings it all inside the @tf while

src/ops/nn.jl
```diff
@@ -137,51 +137,65 @@ function rnn(cell, inputs; initial_state=nothing, dtype=nothing, sequence_length
...
+`dynamic_rnn(cell, inputs, sequence_length, initial_state=nothing, dtype=nothing, parallel_iterations=nothing, swap_memory=false, time_major=false, scope="RNN")`
 
 Creates a *dynamic* recurrent neural network. Performs full dynamic unrolling of `inputs`.
 
 Args:
 * `cell`: An instance of `RNNCell`.
 * `inputs`: A `Tensor` of shape `[max_time, batch_size, ..., ...]` or `[batch_size, max_time, ..., ...]` (see `time_major`). May also be a nested `Tuple` with the same property. The first two dimensions *must* be the same across
 all elements but all later dimensions may vary.
-* `sequence_length`: Specifies length of each sequence in `inputs`.
+* `sequence_length`: Specifies length of each sequence in `inputs`. se
```
malmaud: Is this trailing 'se' a typo?

src/ops/nn.jl
```diff
@@ -137,51 +137,65 @@ function rnn(cell, inputs; initial_state=nothing, dtype=nothing, sequence_length
...
-@op function dynamic_rnn(cell, inputs; sequence_length=nothing, initial_state=nothing, dtype=nothing, parallel_iterations=nothing, swap_memory=false, time_major=false, scope="RNN")
-    sequence_length === nothing || error("sequence_length parameter not supported yet")
-    time_major == false || error("Time-major order not supported yet")
-    time_step = tf.constant(1)
-    num_steps = convert(Tensor{Int64}, tf.shape(inputs)[2])
+@op function dynamic_rnn(cell, inputs, sequence_length=nothing; initial_state=nothing, dtype=nothing, parallel_iterations=nothing, swap_memory=false, time_major=false, scope="RNN")
+    #TODO Make this all work with non-3D inputs
+
+    if time_major
+        inputs=permutedims([2,1,3])
```
malmaud: `permutedims(inputs, ...)`

src/ops/nn.jl
```diff
@@ -137,51 +137,65 @@ function rnn(cell, inputs; initial_state=nothing, dtype=nothing, sequence_length
...
         data = tf.squeeze(data, [2])
+        #TODO: Once indexing supports it, the whole of the above can be replaced with: `data=inputs[:, timestep, :]`
         local new_state
+        new_output = output
+
+
         variable_scope(scope) do
-            output, new_state = cell(data, state, input_dim)
+            new_output, new_state = cell(data, state, input_dim)
+            if !isa(sequence_length, Void) #This should be removed by the julia lowering process
```
malmaud: I'd prefer `sequence_length !== nothing`

src/ops/nn.jl
```diff
@@ -137,51 +137,65 @@ function rnn(cell, inputs; initial_state=nothing, dtype=nothing, sequence_length
...
+                #Only update output and state for rows that are not yet passed their ends
+                have_passed_end = sequence_length .< time_step
```
malmaud: Can we use `<` instead of `.<`? These are all scalars, right?

test/nn.jl
```diff
@@ -1,32 +1,59 @@
...
+        #TODO uncomment out Flaky tests
+        #run(sess, global_variables_initializer()) #This line is flakily breaking on Travis
+        #output = run(sess, y)[1]
+        #expected_output = tanh(1*.1+tanh(1*.1+.1)*.1+.1)
+        #@test output[1,1] ≈ expected_output
+    end
+end
+
+"Dummy cell for testing, always gives back its input"
+immutable IdentityRNNCell <: nn.rnn_cell.RNNCell
```
malmaud: This is a really good idea. Might as well move it to rnn_cell.jl so it can easily be used for other tests and pedagogical purposes.

src/ops/nn.jl
oxinabox: I should have written a test-case for this. My bad, there.

src/ops/nn.jl
oxinabox: looks like

src/ops/nn.jl
oxinabox: So the trick here though, is  I want to be sure that the JIT/specialization/lowering
removes this branch.
And specialization is based 
I know that specialization works with `if !isa`, as that is a type thing (and I tested).
I don't know that it works with `if !==`, as that is a value thing.
I will run tests and report back if it don't work.

src/ops/nn.jl
oxinabox: `time_step` is a scalar.

`sequence_length` is normally a vector -- you get to specify 1 per element in the batch.

It can be a scalar, but this is not the most effective code for that case -- instead in that case should take a slice before starting the `dynamic_rnn`.

src/ops/nn.jl
oxinabox: Ok, seems to be working with `if !==`  so I will make the change

oxinabox:
As am extra bonus.
I made similar changes to the static `rnn`.

Because the dynamic_rnn is kinda broken in a few ways; as you mentioned

PR #189, opened 2017-03-25
Check the sizes of the inputs are compatible with the shapes of the placeholders they fill

oxinabox:
No-one would put the wrong size into a placeholder intensionly.
And we already have mechanisms to allow parts of a placeholder to have flexible sizes.

Putting the wrong sized input can render uses of shape inference incorrect.
So we should not let people do it.

### DIFF
--- a/examples/logistic.jl
+++ b/examples/logistic.jl
@@ -20,22 +20,22 @@ y = draw(y_prob)
 
 # Build the model
 sess = Session(Graph())
-@tf begin
-    X = placeholder(Float64, shape=[-1, 50])
-    Y_obs = placeholder(Float64, shape=[-1, 10])
 
-    variable_scope("logisitic_model") do
-        global W = get_variable([50, 10], Float64)
-        global B = get_variable([10], Float64)
-    end
+X = placeholder(Float64, shape=[-1, 50])
+Y_obs = placeholder(Float64, shape=[-1, 10])
 
-    Y=nn.softmax(X*W + B)
-
-    Loss = -reduce_sum(log(Y).*Y_obs)
-    optimizer = train.AdamOptimizer()
-    minimize_op = train.minimize(optimizer, Loss)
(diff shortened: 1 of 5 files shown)


src/core.jl
```diff
@@ -151,10 +151,16 @@ function TensorShape(dims::Vector)
     TensorShape([x<0 ? Nullable{Int64}() : Nullable{Int64}(x) for x in dims])
 end
 
-function TensorShape(dim::Void)
+function TensorShape(::typeof(collect(tuple())))
```
malmaud: What does this complicated type mean? Is there is a more compact type signature that would work?

examples/logistic.jl
```diff
@@ -19,20 +20,22 @@ y = draw(y_prob)
 
 # Build the model
 sess = Session(Graph())
-X = placeholder(Float64, shape=[nothing, 10])
-Y_obs = placeholder(Float64, shape=[100])
+@tf begin
```
malmaud: Can we make this change to the example a separate PR?

src/core.jl
oxinabox: Yes, could instead say: `Vector{Union{}}`.

I was playing with the idea, that `(::typeof(collect(tuple())` might be clearer.
Since it does not require remembering that collecting some untyped (not Any-typed or Void typed) collection gives `Vector{Union{}}`.

I will make the change

examples/logistic.jl
oxinabox: The example does need to change, as it uses inconstant sizes; and is a test, which would otherwise fail.

But I can remove the `@tf` because I only added that to help debug, which line had the inconstant sizing.

oxinabox: done

src/core.jl
oxinabox: done

Issue #196, opened 2017-03-29
Better Graph Def seems to break everything for me.


oxinabox:
https://github.com/malmaud/TensorFlow.jl/commit/e71d1c88ff0fdcab09520260ffbfa07fbed1a78e

Not sure why it is passing Travis,
but now, anything that was defined `@with_graph_def`,
that  does not take kwargs is broken,
and gives an error: Eg:

>function add_to_collection does not accept keyword arguments
 in kwfunc(::Any) at boot.jl:236
 in #Variable#62(::String, ::Bool, ::Bool, ::Type{T}, ::Array{Float64,2}) at variable.jl:46
 in (::Core.#kw#Type)(::Array{Any,1}, ::Type{TensorFlow.Variable}, ::Array{Float64,2}) at <missing>:0
 in #get_variable#91(::Bool, ::Array{Any,1}, ::Function, ::String, ::Array{Int64,1}, ::Type{T}) at variable.jl:249
 in (::##1#2)() at logistic.jl:28
 in #variable_scope#90(::Array{Any,1}, ::Function, ::##1#2, ::String) at variable.jl:208
 in (::TensorFlow.#kw##variable_scope)(::Array{Any,1}, ::TensorFlow.#variable_scope, ::Function, ::String) at <missing>:0
 in include_string(::String, ::String) at loading.jl:441
 in include_string(::String, ::String, ::Int64) at eval.jl:28
 in include_string(::Module, ::String, ::String, ::Int64, ::Vararg{Int64,N}) at eval.jl:32
 in (::Atom.##53#56{String,Int64,String})() at eval.jl:40
 in withpath(::Atom.##53#56{String,Int64,String}, ::String) at utils.jl:30
 in withpath(::Function, ::String) at eval.jl:46
 in macro expansion at eval.jl:57 [inlined]
 in (::Atom.##52#55{Dict{String,Any}})() at task.jl:60

This one comes when I run the example logistic code.

Version Info
```
Julia Version 0.5.0
Commit 3c9d753 (2016-09-19 18:14 UTC)
Platform Info:
  System: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-5930K CPU @ 3.50GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
[...]

PR #197, opened 2017-03-29
solve #196 need julia >= v0.5.1

oxinabox:
This solves #196 
Make sure no-one is on julia 0.5.0

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.5
+julia 0.5.1
 Requests 0.3.10
 BinDeps 0.4.4
 ProtoBuf 0.3.0


malmaud:
Oh hmm, I don't really want to leave the .5 users in the cold over such a small thing. Why don't we just have an if statement that loads the old version of get_def_graph if users are `<0.5.1`, and eventually we'll delete that block?

oxinabox:
0.5.0 -> 0.5.1 is a really trivial upgrade.
It took me 2 minutes to download the newer release, for linux.

It is not like 0.5.0 is in package managers, and 0.5.1 is not.
0.4.5 is in debian's repo.

I believe @staticfloat's PPA has 0.5.1,
(https://launchpad.net/~staticfloat/+archive/ubuntu/juliareleases/)

and in Homebrew is is 0.5.1 for sure https://github.com/staticfloat/homebrew-julia

Anyone who needs 0.5.0 can just stay with using the last tagged Tensorflow release,
prior to the changes to `@with_def_graph`
It is pretty functional.
And 0.6 is just around the corner.

Alternatively, I think it can be changed in a way that works for 0.5.0 and 0.5.1 easily enough.
But I don't care to have two different patch versions of julia installed to find it.
(I'm thinking it is to delete the semicolon from the call, when there are no keyword-arguments. But not tested that)

malmaud:
We will eventually have a branch that only works on Julia 0.6, but I at least want the next tagged version of TensorFlow to still work on .5.0. 

Some people have Julia installed by system administrators on clusters or might otherwise be annoyed to have to upgrade Julia to .5.1, even if (as I agree) it's simple, so I'm still reluctant to require people to upgrade to .5.1.

malmaud:
I'm going to close in favor of https://github.com/malmaud/TensorFlow.jl/commit/3d8b91c76b14e97e1282172b2db1966b52277602 for now, but we can delete this once we have a Julia .6 branch.

oxinabox:
fair enough.

Issue #187, opened 2017-03-25
Make `[a; b]` and `[a b]` concatencation work


oxinabox:
I realised that it is possible to overload julia's matrix concatenation syntax.
This is done by overloading `hcat`, `vcat`. And I think optional `hvcat`.
Like `end` is changed to `endof`,  `[a; b]` is changed to `vcat(a,b)`, and `[a b]` is changed to `hcat(a,b)`
And mixes of the two are changed to `hvcat`.

So we can overload these to call the appropriate TensorFlow `Concat` operations.
Is this a good idea?

For a useless example, showing the overload can be done:

```julia

immutable Foo
    x
end


#Show Vcat can overload
Base.vcat(fs::Vararg{Foo}) = sum(f->f.x, fs)
@assert [Foo(1); Foo(2)] == 3

#Show hcat can overload
Base.hcat(fs::Vararg{Foo}) = prod(f->f.x, fs)
@assert [Foo(1) Foo(2)] == 2


# Make sure not broken commas in array expressions anything else.
@assert [Foo(1), Foo(2)][1]==Foo(1)
@assert [Foo(1), Foo(2)][2]==Foo(2)
```

PR #209, opened 2017-04-28
[A; B] and [A B] concatenation syntax (#187)

oxinabox:
does #187
So that cool.

This could do with some code review (like all code).
It contains a `@goto` which is frowned upon by Dijkstra.
Normally I want to use gotos to break out of nested loops, and then I immediately see that I want to refactor to be using a function and then can use return to breakout.

But this time I want to restart the loop from the beginning
One option would be to use a while loop and manage the index manually
But I don't think that is more readable.

One thing that is still pending is a general way for this to be dispatched to when one or more of the varargs is an `AbstractTensor` and one or more is not (but assumably can be converted).
Right now this is accomplished by specialcasing the first two arguments.
So long as one of the first two arguments is an `AbstractTensor` then it is dispatched to correctly.
I think that covers the vast majority of uses.

Maybe it is worth expanding to first 3? Beyond that doing it manually becomes cumbersome.
I could perhaps workout a metaprogramming loop and do it to the first dozen or so.

----------

Not sure what to do with `hvcat` yet.
I don't really know where it is used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ tfdocs
 docs/build
 *.DS_Store
 *.*~
+
+*.swp
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ run(sess, global_variables_initializer())
 checkpoint_path = mktempdir()
 info("Checkpoint files saved in $checkpoint_path")
 for epoch in 1:100
-    cur_loss, _ = run(sess, vcat(Loss, minimize_op), Dict(X=>x, Y_obs=>y))
+    cur_loss, _ = run(sess, [Loss, minimize_op], Dict(X=>x, Y_obs=>y))
     println(@sprintf("Current loss is %.2f.", cur_loss))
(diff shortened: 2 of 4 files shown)


oxinabox:
TODO: Readme needs to be updated, to not use `vcat` in examples
Because that is taken now.  but to use `Vector` instead, or to to just write `[a, b]`, which is prob cleanest

oxinabox:
TODO:
check that `vcat(foo)==foo` and `hcat(foo)==foo'`

I suspect they are errors right now.

oxinabox:
With that corner case closed,
I would very much like this to be reviewed and merged, (@malmaud  @MikeInnes)
and ideally a new Tensorflow.jl version tagged.

We are using this syntax part in **[TensorFlowDiffEq.jl](https://github.com/JuliaDiffEq/TensorFlowDiffEq.jl/blob/157209cd21aef12d1414fbaa2aa7cc9f8cb8c856/test/runtests.jl#L51)** (@ChrisRackauckas)

And it is hard to deploy tests against non-tagged versions.

PR #210, opened 2017-04-28
global_variables_initializer -> initialize_all_variables

mspiffy:


malmaud:
The usage is actually about to change to be `global_variables_initializer`

oxinabox:
This has always been an issue with Readme's on github.
Since github shows the default branch's readme as  on the project home page.
And the default branch is _master_.

The readme is correct for _master_.

Possible general solutions:

 - change the default branch in settings
 - Have masters have a second readme file called `README.current_branch.md`, and have `README.md` used only for the last release -- then do an update per release.
 - Not have any examples in the readme, and have a link to Docs-Release, and Docs-latest

Issue #212, opened 2017-05-01
Extending the `gradients` function


MikeInnes:
The `gradients` function currently supports neither multiple `y` values nor providing the gradients of those values.

It hopefully shouldn't be too hard to add this, I think it's just a case of forwarding those arguments correctly to the Python side (but please correct me if I'm wrong there).

oxinabox:
Can you clarify what you mean by "providing the gradients of those values" ? 
Do you mean providing some gradients as inputs to the function?

MikeInnes:
Yes, in any case where the computation produces more than a single scalar output, `gradients` needs to know the gradients of those outputs to produce a meaningful result.

```julia
julia> X = Variable([1,2,3])
TensorFlow.Variable(<Operation 'node_3'>, <Operation 'node_3/Assign'>)

julia> Y = Variable([4,5,6])
TensorFlow.Variable(<Operation 'node_4'>, <Operation 'node_4/Assign'>)

julia> Z = X+Y
<Tensor add_2:1 shape=(3) dtype=Int64>

julia> run(sess, global_variables_initializer())

julia> run(sess, gradients(Z, [X, Y]))
2-element Array{Array{Int64,1},1}:
 [1, 1, 1]
 [1, 1, 1]
```

This produces a result, but the gradients default to arrays of `1`s.

oxinabox:
I'll take a poke at this soonish, maybe sometime this week.
You are right that it should be pretty simple.

I think I want it anyway for some of the DiffEq stuff I am talking to @ChrisRackauckas about.

PR #215, opened 2017-05-10
Allow more options for Gradient calculation

oxinabox:
Solves #212

Possible issue with implementation `node_name` is now applied element-wise to `AbstractVectors`.
It makes the implementation short and simple but I'm not sure how I feel about it

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -1370,16 +1370,19 @@ Base.haskey(graph::Graph, name) = isnull(get_node_by_name(graph, name))
 
 
 
-node_name(xs::AbstractVector) = node_name.(xs)
+node_name(::Void) = nothing
+node_name(xs::AbstractVector)=node_name.(xs)
 
-function gradients(y, x::AbstractArray)
+
+function gradients(y, x::AbstractArray, grad_y=nothing)
     x_names = node_name(x)
     y_names = node_name(y)
+    grad_y_names = node_name(grad_y)
     meta_graph = train.export_meta_graph()
     b = IOBuffer()
     writeproto(b, meta_graph)
     graph_proto = @compat take!(b)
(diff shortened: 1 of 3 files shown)


oxinabox:
Not sure why coverage is decreasing.
Is py.jl getting skipped because it is another process?

src/core.jl
```diff
@@ -1370,15 +1370,37 @@ Base.haskey(graph::Graph, name) = isnull(get_node_by_name(graph, name))
 
 
 
+node_name(::Void) = nothing
+node_name(xs::AbstractVector)=node_name.(xs)
```
MikeInnes: Couldn't you just call `node_name.(x)` whereever you expect an array? On 0.6 at least that should behave the right way for a single tensor.

oxinabox: Can you? I'll test this.

malmaud:
Ya sure, thanks. I got busy with end of semester things, although I am
still plugging away at my autoops branch (it's a pretty big change).

src/core.jl
oxinabox: Tested it, doesn't work out.
Because we don't always know when we expect an array.
`gradients(y, x, grad_y)`
can have any of its 3 parameters as an array of Tensors or as a Tensor, (though if `y` is an array `y_grad` must also).

And on 0.5 at least `nodename.(x)` for `x::Tensor` comes back with: 
```julia
  MethodError: no method matching start(::TensorFlow.Tensor{Int32})
```

So with that trialed, I will merge this now, as is.

MikeInnes: Yes, this would only work on 0.6.

It may be a good idea to drop 0.5 fairly quickly, but that's up to you / Jon.

PR #216, opened 2017-05-17
WIP: Automatically import all TensorFlow ops

malmaud:


src/ops.jl
```diff
@@ -156,6 +157,253 @@ function struct_map(op, args...)
...
+        module Ops
+        import TensorFlow
+        const tf = TensorFlow
+        """)
+        for (name, op) in get_all_op_list()
+            try
+                f = to_function(op)
+                s = stringify_func(f)
+                write(ops_file, s)
+                Base.print(ops_file, "\n\n")
```
oxinabox: Why is that `Base.print` and not just `print` ?

oxinabox:
Looking nice. I have been watching the branch, casually.

Do also set the .gitattribute for generated code:
https://github.com/github/linguist#generated-code

malmaud:
@oxinabox , if you get  a chance, want to do a reviewing pass?

oxinabox:
Yes,
I'll try and do it in the next few days, if that is good with you.

So this is g2g, otherwise?

malmaud:
Yes, except for a small issue with .5 that I'll fix soon.

oxinabox:
I suggest excluding generated files from Codecov too
https://docs.codecov.io/v4.3.6/docs/ignoring-paths

Since so long as the functions are generated correctly (which is tested) they should be correct.

oxinabox:
Did you mean to merge that?
I was going to do the review tomorrow ...

malmaud:
Oh sorry, I was going out of town soon and wanted to tag a new version before the weekend. But I'd still welcome comments against `master`.

oxinabox:
Ok.
Well I'll do that tomorrow then. 
It is quiet late in Australia.
g'night

oxinabox:
I think I can still create a Review on the merged PR,
which is probably easiest way to organize the information.
Then can make another PR for any changes needed.

malmaud:
Ah, that's a good idea.

test/clipping.jl
```diff
@@ -5,7 +5,7 @@ sess = TensorFlow.Session(TensorFlow.Graph())
 a_raw = rand(10, 10)
```
oxinabox: These tests need a bit of tidying.
Not related to this PR,
but at some point I will go through and give them `testset`s and remove excess `TensorFlow.foo` module names.

test/core.jl
```diff
@@ -55,34 +55,10 @@ end
...
-
-    # Should get back output
-    x_val = rand(100, 50)
-    pred_oh = run(sess, s["Y_pred_onehot"], Dict(s["X"] => x_val))
-    @test size(pred_oh) == (100, 10)
-
-    # Should be able to use output for math still
-    pred = run(sess, indmax(s["Y_pred_onehot"], 2), Dict(s["X"] => x_val))
-    @test length(pred) == 100
+    g = Graph()
```
oxinabox: Why have these been deleted?
Have equiv tests been added elsewhere?

test/summary.jl
```diff
@@ -5,13 +5,15 @@ using Base.Test
...
     x_summary = tf.summary.scalar(["x", "y"], x)
     hist_summary = tf.summary.histogram("z", randn(10))
     summaries = tf.summary.merge_all()
     summary_pb = run(sess, summaries)
-    @test_nowarn write(writer, get_def_graph())
-    @test_nowarn write(writer, summary_pb)
-    @test_nowarn close(writer)
+    write(writer, get_def_graph())
+    write(writer, summary_pb)
+    close(writer)
```
oxinabox: why not testing for `nowarn` anymore?

src/ops/nn.jl
```diff
@@ -1,119 +1,52 @@
...
-sampled_softmax_loss,
-batch_normalization,
-seq2seq,
-conv2d_transpose
-
 using Compat
-import ..TensorFlow: Operation, NodeDescription, get_def_graph, capitalize, add_input, Port, get_name, set_attr_list, get_shape, variable_scope, shape, random_uniform, AbstractTensor, Tensor, reduce_sum, @not_implemented, with_op_name, @op, @tf
 import TensorFlow
 const tf = TensorFlow
+import .tf: Ops, @op, @not_implemented, @tf
```
oxinabox: What is the reason for cutting down on imports for parent module?

I was thinking yesterday we should do the opposite and just say `using ..TensorFlow`

README.md
```diff
@@ -55,6 +55,8 @@ Pkg.build("TensorFlow")
 CUDA 8.0 and cudnn are required for GPU usage.
 If you need to use a different version of CUDA you can [compile libtensorflow from source](#optional-building-the-tensorflow-library).
 
+Initial precompilation (eg, the first time you type `using TensorFlow`) can take 10-15 minutes, so please be patient. Subsequent load times will only be a few seconds.
```
oxinabox: Took well under 5 minutes on my computer, so maybe say 5-15 minutes,
Or up to 15 minutes (but often much less).
So people don't freak out.

oxinabox:
Is there a pressing reason to `imported_ops.jl` into version control?
The other option is to generate during `Pkg.build`.
This has the advantage of ensuring it is insync with the version of `libtensorflow.so`.
It would increase `Pkg.build` time (idk how much -- depends on CPU speed)
But it would decrease `Pkg.add` time (idk how much -- depends on internet speed).

Given out `Pkg.add` and `Pkg.build` times are both long already, I doubt time would make much of a different.
So I guess the primary concern is keeping it insync with `libtensorflow.so`

src/ops/sequences.jl
```diff
@@ -1,244 +1,53 @@
...
-* `shape`: A one dimensional `Tensor` or array containing the shape of the output `Tensor`.
-* `minval`: A zero dimensional `Tensor` or value of type `dtype`. Lower bound on random values.
-* `maxval`: A zero dimensional `Tensor` or value of type `dtype`. Upper bound on random values.
-* `seed`: An integer to seed the RNG with. Defaults to `0`.
-* `dtype`: Optional datatype of random values generated. Default is `Float32`.
-
-Returns:
-A `Tensor` of the specified `shape` and `dtype` containing random values.
-"""
-@op function random_uniform(shape, minval, maxval; name=nothing, seed=0, dtype=Float32)
```
oxinabox: This change broke `nn.dropout`
The new imported op has a different signature:
and does not support minval, maxval or dtype.
`random_uniform(shape; seed=0, seed2=0)`

I'll make a PR shortly to bring this back,

oxinabox: See #233

src/ops.jl
```diff
@@ -6,20 +6,40 @@ const tf = TensorFlow # so know where op_funcs is defined
 using MacroTools
 
 if VERSION < v"0.6.0-"
-    import Base: .*, .+, ./, .-, .^, .==
+    import Base: .*, .+, ./, .-, .^, .==, .!=
 end
 
-
-
-function tf_promote(t, x::Number)
```
oxinabox: Are you sure we don't want this?
This is a surprisingly breaking change.
Because it is what makes writing `- 1` work on `Tensor{Int32}`s to get back an `Tensor{Int32}` and not a `Tensor{Int64}`

This does being us inline with the change that was made from 0.5 to 0.6
https://discourse.julialang.org/t/is-this-expected-behavior-0-5-to-0-6-breakage/2534/7

This explains the extra convert calls I noted in in the Imported_ops generation code.

src/ops/transformations.jl
```diff
@@ -52,58 +37,14 @@ https://www.tensorflow.org/versions/r0.10/api_docs/python/array_ops.html#reshape
...
-        add_input(desc, n)
-        add_input(desc, dims)
-    end
-    Tensor(Operation(desc), 1)
+@op function Base.reshape(n::AbstractTensor, dims::AbstractTensor; kwargs...)
+    Ops.reshape(n, dims; kwargs...)
 end
 
 
 # if isdefined(Base, :slice)  # Removed in .6
```
oxinabox: This is a breaking change for users of `slice` in 0.5,
since it was effectively exported before (by overloading `Base`)
and it is not exported now.

It is exported in 0.6
https://github.com/malmaud/TensorFlow.jl/blob/master/src/TensorFlow.jl#L128

PR #217, opened 2017-05-19
Add testing on 0.6

staticfloat:


malmaud:
We shouldn't merge this until the tests actually pass on .6. I'm working on it.

oxinabox:
We could merge it,
then it would be clear *when* the tests pass.
and that they don't currently.

I was thinking about fixing it for 0.6. (though that would mean me getting a copy of 0.6)
But some of the fixes are already maybe done in your branch.
Eg:
https://github.com/malmaud/TensorFlow.jl/pull/216/commits/8ce620f739d47ef7a5275c72bbf0cd2c4c382fb0#diff-7108b1394b2df366f0da990483e66145R13

malmaud:
It should be put in the `allow_failures` section then.

PR #223, opened 2017-05-24
Document default graph changing functions

oxinabox:
I can never remember the names for these.
So I documented them.
and included cross references.

I did not include references to `set_def_graph` as that is not exported.
Presumably because we would rather people used `as_default`

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -209,7 +209,7 @@ function Base.show(io::IO, g::Graph)
     print(io, "Graph($(g.ptr))")
 end
 
-# There is a bug in Julia v0.5 that requires using the
+# There is a bug in Julia v0.5.0 that requires using the
 # the non-MacroTools version of this function. The old
 # branch will be eliminated eventually.
 if VERSION >= v"0.5.1"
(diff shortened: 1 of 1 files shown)


src/core.jl
```diff
@@ -429,10 +431,31 @@ function get_def_graph()
 end
 has_def_graph() = isdefined(def_graph, :x)
 
+"""
+Sets the default computation graph to `g`.
```
malmaud: Can you include the function signature at the top of the docstring, as seemed to be conventional?

src/core.jl
```diff
@@ -429,10 +431,31 @@ function get_def_graph()
...
+
+See also `get_def_graph`, `as_default`
+"""
 function set_def_graph(g)
     def_graph[] = g
 end
 
+"""
+For the duration of the function `f`
```
malmaud: And here

src/core.jl
oxinabox: Sure,
I wonder if we can use DocumenterExtra.jl to do this?

oxinabox:
Fixed the doc strings.
And a few extra while I was there.

oxinabox:
@malmaud while you are merging PRs.
Don't forget this one. :-D

malmaud:
Ah thanks

PR #227, opened 2017-05-26
Make RNNs not need batch size to be statically known

oxinabox:
The main thrust of this PR is to allow the batch_size of a RNN input to be able to be unknown.
So it can be 

for a static RNN eg
```
data = placeholder(Float32; shape=(-1,20, 25)
```

And for a dynamic RNN eg:
```
data = placeholder(Float32; shape=(-1,-1, 25)
```

The PR does a few other things/

 - I found that time-major actually didn't work as a parameter (Might need to add a test for that still, though I have checked it manually)
 - It was too easy to do illegal sliced indexing and get stackoverflows (see https://github.com/MikeInnes/Flux.jl/pull/28#issuecomment-301220635)
 - TensorShape was messing up on a cornercase that showed up in testing Transpose. So I tweaked it and added tests

This could be broken into 4 PR. for commits 2,3,4,(1+5).
Technically, I think they would each work on their own (maybe not 2 without 3).
But I was solving other things as I solved the main goal.
I could cherry pick-them apart if required.

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -172,6 +172,9 @@ function TensorShape(::Void)
     TensorShape(Nullable{Int}[], true)
 end
 
+function TensorShape(::Vector{Union{}}) # NB: `Vector{Union{}} == typeof(collect(tuple())))`
+    TensorShape(Nullable{Int}[], false)
+end
 
 function get_shape end
 
--- a/src/core.jl
+++ b/src/core.jl
@@ -159,18 +159,14 @@ end
     rank_unknown::Bool
 end
 
-function TensorShape(dims::Vector{Nullable{Int}})
+
+function TensorShape{T<:Integer}(dims::AbstractVector{T})
+    TensorShape([x<0 ? Nullable{Int}() : Nullable{Int}(x) for x in dims])
+end
+
+function TensorShape(dims)
     TensorShape(dims, false)
 end
-
-function TensorShape(dims::Vector)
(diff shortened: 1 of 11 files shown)


oxinabox:
This does remove the `nn.rnn_cell.zero_state(::RNNCell, batch_size::Int, dtype)`
function.
which I figured would be fine since it is not exported.
And there is no reason anyone would call it, since zero is the default.

but I could put it back, trivially.

oxinabox:
I'm not sure how large the time cost of abusing multiplication to make the dynamically initial state matrix is.

oxinabox:
Facepalm.
There is a dynamic `fill` operation.
https://www.tensorflow.org/api_docs/cc/class/tensorflow/ops/fill

That will be much better than my cunning multiplication hack

oxinabox:
@malmaud this is ready for review.

It took a bit to get it going again after rebasing on top the changes from importing all ops had to be dealt with. But those PRs are merged now.
And the sanity of using `fill` is added.

And I've now tested it on my research code extensively.
Which seems a really solid test, since it broke heavily for other things. 
I guess I use extreme ops in some way.

oxinabox:
@malmaud I have rebased this, again.
Any timeline on when you will be able to review?

malmaud:
Sometime in the next few days.

malmaud:
This looks great, thanks.

PR #231, opened 2017-05-27
be more specific about VERSION cutoffs in ops.jl

tkelman:
avoid breaking the package on early dev versions, in case bisecting
on base is ever needed to isolate a bug triggered by this package


malmaud:
Thanks

PR #232, opened 2017-05-29
Add test for size and xor

kshyatt:
I think we don't have tests for `logical_xor` at all or the op symbol -> function thing.

### DIFF
--- a/test/hello.jl
+++ b/test/hello.jl
@@ -21,6 +21,7 @@ result = run(sess, a+b)
 #basic tensor facts
 tens = TensorFlow.ones((5,5))
 @test size(tens) == (5,5)
+@test size(tens, 1) == 5
 @test length(tens) == 25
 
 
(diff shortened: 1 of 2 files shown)


malmaud:
Thanks! Looks like these don't pass on .5, though.

test/hello.jl
```diff
@@ -21,6 +21,7 @@ result = run(sess, a+b)
 #basic tensor facts
 tens = TensorFlow.ones((5,5))
```
oxinabox: Isn't this just calling `Base.ones((5,5)` which gives back a Matrix of `Int` ?

Should this be `ones(Tensor, (5,5))` ?
And if so then `size` and `length` won't return `Int`s, they will return `Tensor{Int}`s,
because `size(::Tensor)`  is an overload for `shape(::Tensor)`, not `get_shape(::Tensor)`

malmaud:
I'm gonna close since this is pretty outdated by now.

PR #233, opened 2017-05-30
Fix the uniform_random used by dropout

oxinabox:
>The new imported op has a different signature:
and does not support minval, maxval or dtype.
random_uniform(shape; seed=0, seed2=0)

This brings back old-functionality. So that `dropout` isn't broken anymore.

I'm not sure what is meant in the TODO about using global seed.
Does that mean something like 
`seed1=Base.Random.GLOBAL_RNG.seed |> first` ?


malmaud:
Thanks

PR #234, opened 2017-05-30
Clean up generated code

MikeInnes:
This cleans up the generated code a bit, which significantly improves the amount of output (by a third) – the load time is improved by the same amount.


src/ops.jl
```diff
@@ -219,8 +217,8 @@ function to_function(op::tensorflow.OpDef)
     input_block = quote end
     convert_block = quote end
     type_sets = Dict{String, Vector{Symbol}}()
-    for input in op.input_arg
-        sym = gensym()
+    for (i, input) in enumerate(op.input_arg)
+        sym = Symbol(:x, i)
```
oxinabox: I feel like this can be `Symbol(op.input_arg.name)`.
Which will match the docs.
But possibly that would make a name clash?
If so appending a underscore or similar might be the trick.

MikeInnes: Yeah, that would make sense.

MikeInnes:
As far of the broader design of this is concerned, I suggest having a generic `to_op(::OpDef, x1, x2...)` function that does checks and conversions for any possible op. Then each generated function would just be a docstring and a call like `@op asin(x) = to_op($opdef, x)`, which would be significantly easier on Julia's compiler. (You'd also want to deserialise a code block rather than parsing text.)

Right now this file takes about 0.5s per function definition to load, which seems suspiciously high. Perhaps `@op` is causing the bottleneck here?

oxinabox:
So you cut down the number of lines by about 45%, and the code is a bit prettier without so many begins.

And you got a loading speedup?

MikeInnes:
Yup, from about 300 to 200 seconds for me.

src/ops.jl
malmaud: Good idea, I made that change.

malmaud:
Thanks Mike, very useful.

Issue #218, opened 2017-05-21
Feature request: Wrap op tf.svd


alha02:

oxinabox:
Are you in any kinda rush? This will be done when #216 is done.

Doing it by hand is easy enough though.
See https://github.com/malmaud/TensorFlow.jl/blob/master/docs/src/addingops.md

alha02:
Oh, sorry I didn't realize that this was in the pipeline.

oxinabox:
Reopening this, because it wasn't quiet closed by #216.

svd is there in the Ops submodule,
but it needs a bit of turning around to be made to overload `Base.svd`.
Just reordering the returned outputs, I think

malmaud:
Closed in https://github.com/malmaud/TensorFlow.jl/commit/12e71f018d8ec8b43afed3cf9f214d4c9ad6069b.

PR #240, opened 2017-05-31
add SVD binding re: #218

oxinabox:
This wraps up the 

I am surprised at how poor the reconstruction error from the Tensorflow result is
I  found in one test 
 - the  julia error at 4.346551f-12
 - and the tensorflow error at 1.1636697f-10
Thus the factor of 100 in checking the error value.

Julia just [wraps LAPACK `gesdd!`]https://github.com/JuliaLang/julia/blob/bc606f6cc8bc6127979d90c6ccb239b2e2a14448/base/linalg/svd.jl#L28)

I am going to investigate further if there is actually a difference in accuracy of it it was just some bad luck for a particularly ill conditioned matrix in some sense.
But if I find anything wrong it is an upstream issue, anyway.

@alha02

### DIFF
--- a/src/ops/math.jl
+++ b/src/ops/math.jl
@@ -30,6 +30,15 @@ import .Ops:
 @op Base.max(x::AbstractTensor, y; kwargs...) = Ops.maximum(x, y; kwargs...)
 @op Base.min(x::AbstractTensor, y; kwargs...) = Ops.minimum(x, y; kwargs...)
 
+
+@op function Base.svd(a::AbstractTensor; thin=true, kwargs...)
+    # Match Base names and ordering of results
+    s,u,v = Ops.svd(a; compute_uv=true, full_matrices=!thin, kwargs...)
+    u,s,v
+end
+
+
+
 const multiply = Ops.mul
 const negative = Ops.neg
 const self_adjoint_eig = Ops.self_adjoint_eig_v2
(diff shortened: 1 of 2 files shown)


malmaud:
sweet

oxinabox:
For interest:
I completed some evalutations
http://white.ucc.asn.au/2017/06/02/TensorFlow's-SVD-is-significantly-worse-than-LAPACK's,-but-still-very-good.html

TensorFlow's SVD *is* bad compared to base Julia's.
But I think it is still so good that it doesn't matter to anyone.

jiahao:
I think the discrepancy comes down to different logic in the SVD algorithms. TensorFlow [uses](https://github.com/tensorflow/tensorflow/blob/754048a0453a04a761e112ae5d99c149eb9910dd/third_party/eigen3/Eigen/SVD) Eigen's algorithm, which is Gu-Eisenstat divide and conquer with QR-preconditioned Jacobi rotations for matrix blocks of 16x16 or smaller (governed by `m_algoswap` on L107 [here](https://bitbucket.org/eigen/eigen/src/f25fd30ccc18be172b2da75b31ec61b3161c2c93/Eigen/src/SVD/BDCSVD.h) and the switchover is on L247). If only the singular values are requested, LAPACK doesn't use Gu-Eisenstat divide and conquer at all, but rather bulge chasing (in DLASDQ) as described in LAWNS 3 [(pdf)](http://www.netlib.org/lapack/lawnspdf/lawn03.pdf).

PR #243, opened 2017-06-02
fix relative path to generated file in deps/generate_ops.jl

tkelman:
unless you plan on manually moving the file over from deps each time you regenerate it?

edit: may also need an `import MacroTools` in the file?


malmaud:
Ah, thanks. Yep, need `MacroTools` too.

malmaud:
superseded by https://github.com/malmaud/TensorFlow.jl/pull/244

PR #245, opened 2017-06-04
Updates build.jl to use Official TF C Libs

JobJob:
Credit to @asimshankar - I mostly just uncommented Asim's code.

Adds GPU support on MacOS
Updates NEWS and README
More info on lib download location: https://www.tensorflow.org/install/install_c

Only tested on macOS with both `ENV["TF_USE_GPU"] = "0"`, and `ENV["TF_USE_GPU"] = "1"` on julia 0.6-rc1 (n.b. ENV vars must be set in juliarc, ENV values in current REPL aren't used by Pkg.build - I guess it spawns a new julia process)

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -12,10 +12,8 @@ if !isdir(bin_dir)
     run(`mkdir -p $bin_dir`)
 end
 
-# When TensorFlow 1.1 is released, use the official release binaries
-# of the TensorFlow C library. Do this by setting cur_version to 1.1.0
-# and then replacing the blocks below with:
-#=
+const cur_version = "1.1.0"
+
 function try_unzip()
     try
         run(`tar -xvzf $base/downloads/tensorflow.tar.gz --strip-components=2 ./lib/libtensorflow.so`)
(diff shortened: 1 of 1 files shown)


asimshankar:
Sorry to be the bearer of bad news, but starting with 1.2, TensorFlow binary releases are likely to not include distributions for OS X with GPU. We've had trouble keeping the build for Mac+GPU stable given the somewhat non-standard setup for CUDA-enabled cards on MacBooks and MacPros, so will be relying on community support to keep that in shape.

Just giving a heads up.

JobJob:
Right, thanks for the heads up. 

I also just noticed now that my GPU is being ignored with that libtensorflow.so: `Ignoring visible gpu device ... GeForce GT 750M ... with Cuda compute capability 3.0`. So have to compile from source anyway. 

Since this is the NVIDIA card that I imagine most mac users have, I guess I'll remove the updates to the NEWS and README.

Nonetheless the changes in this PR still make sense I think, i.e. to download libtensorflow from `tensorflow/libtensorflow` rather than `malmaud-stuff`.

malmaud:
Thanks! We need https://github.com/tensorflow/tensorflow/issues/8669 to be fixed first though, since my binaries use a custom patch at the moment.

Issue #205, opened 2017-04-13
TFRecordWriter and tf_record_iterator


classner:
Thank you for the great package! I noticed that both, the `TFRecordWriter` and `tf_record_iterator` are currently still unwrapped. They are incredibly useful for creating and reading .tfrecord files. Are you planning to wrap them or are there inherent problems making this impossible?

malmaud:
Just hasn't been a priority. Wouldn't be hard at all.

malmaud:
Also, you're welcome!

PR #248, opened 2017-06-07
Add TF record writer tools

malmaud:
Closes https://github.com/malmaud/TensorFlow.jl/issues/205

### DIFF
--- a/src/TensorFlow.jl
+++ b/src/TensorFlow.jl
@@ -165,13 +165,12 @@ library loaded.
 macro py_proc(expr)
     quote
         eval(Main, quote
-            remotecall_fetch($(TensorFlow.load_python_process())) do
+            remotecall_wait($(TensorFlow.load_python_process())) do
                 $($(Expr(:quote, expr)))
             end
         end)
     end
 end
-
 
 include("meta.jl")
 include("constants.jl")
--- a/src/TensorFlow.jl
+++ b/src/TensorFlow.jl
@@ -161,6 +161,10 @@ end
 
 Run the given code block in the Julia worker with the Python TensorFlow
 library loaded.
+
+Returns a future to the result.
+*Warning*: Calling `fetch` on a result that contains a pointer, such as a
+`PyObject`, will zero-out the pointer. 
 """
 macro py_proc(expr)
     quote
--- a/src/io/tfrecord.jl
+++ b/src/io/tfrecord.jl
@@ -66,7 +66,6 @@ function RecordIterator(path::AbstractString)
 end
 
 function _next(iter::RecordIterator)
-    local record
     try
         record = fetch(@tf.py_proc $(iter.pyo)[:__next__]())
         RecordIteratorState(Nullable(record))
--- a/src/io/tfrecord.jl
+++ b/src/io/tfrecord.jl
@@ -1,0 +1,90 @@
+module TFRecord
+
+export
+RecordWriter,
+RecordIterator
+
+import TensorFlow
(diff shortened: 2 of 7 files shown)


src/io/tfrecord.jl
```diff
@@ -0,0 +1,94 @@
...
+Returns a Julia iterator that returns the records in the TF Record file
+at `path` as `Vector{UInt8}` objects.
+"""
+function RecordIterator(path::AbstractString)
+    pyo = @tf.py_proc py_tf[][:python_io][:tf_record_iterator]($path)
+    RecordIterator(pyo)
+end
+
+function _next(iter::RecordIterator)
+    local record
```
oxinabox: this line does nothing. Does it?

src/TensorFlow.jl
```diff
@@ -165,14 +165,13 @@ library loaded.
 macro py_proc(expr)
     quote
         eval(Main, quote
-            remotecall_fetch($(TensorFlow.load_python_process())) do
+            remotecall_wait($(TensorFlow.load_python_process())) do
```
oxinabox: This can be just `remotecall` I think.
Because eventually the consumer is going to`fetch` it.
And `fetch` always does a `wait` until there is a result

malmaud: The consumer doesn't always fetch the result, but still wants to rest assured the operation was carried out before returning.

oxinabox: Because it is mutating state?
Ok.

So what is the point of doing a `wait` rather than a `fetch` here?
The time taken will be very similar.
Are you just trying to keep the stack-trace nice, for when exceptions are thrown?

malmaud: It's because when a `PyObject` is fetched, the pointer in it is zeroed-out since Julia doesn't want you passing pointers between processes. But you can safely return a future to the pointer with `wait`, which you can then reference in a subsequent call to the Python process.

malmaud:
This is passing for me locally but failing on CI, not sure why.

oxinabox:
>This is passing for me locally but failing on CI, not sure why.

When I check this branch out on one of my machines it fails in the same way as CI.
I'm now taking a quick look.

src/TensorFlow.jl
oxinabox: Ah I see.

oxinabox:
Got it.
I think it should now be working everywhere,
if my guesss that you have PyCall hitting python , is correct.

It is a python 2 vs python 3 difference: https://www.python.org/dev/peps/pep-3114/

src/TensorFlow.jl
oxinabox: For my reference  why is the python stuff it it's own process?
It is  not being used for speed (since there is a wait).
And it isn't improving stability either.
I've crashed the python hosting worker several times.
Normally it takes out that process and never recovers.
Sometimes the crash immediately cascades into a segfault taking out the main process as well.
(or a OOM)

I guess if it takes out the worker process, user-code could be written to softly crash.
or in the future code could be added to TensorFlow.jl, to reboot the worker once or twice to see if it were a transient fault.

malmaud: It's because Python loads its own version of libtensorflow in the main namespace, which conflicts with TensorFlow.jl's use of libtensorflow. It crashed when I tried and it seems to be a known issue. It would be great to find a solution that didn't involve using a second namespace.

malmaud:
Ah, didn't think of the python 2 vs 3 thing, great catch and fix.

PR #253, opened 2017-06-09
Reduce the number of dependancies

oxinabox:
The number of dependencies we have is huge. (see https://juliaobserver.com/packages/TensorFlow for upstream)
This cuts it down a little.
Biggest change being to get rid of Requests.jl, and just use `Base.download` in the build.jl.
This should knock out maybe 5 or more further upstream requirements.

This also gets rid of a couple of other requirements, that are not used at all.

Other packages maybe that could be shed:
 -  Juno.jl
     - only used for a small amount of improved `show`
     - this would also get rid of the indirect Hiccup.jl and Media.jl, requirement
     - I think we should keep this; hopefully optional packages will come in soon (lots of discussion going around there)
 - Distributions.jl
    - This would knock out about 8 upstream dependencies
    - I don't think we do much with it that couldn't be done with `Base.rand` and `Base.randn` 
    - Probably worth its own PR, if it is worth doing

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -8,3 +8,4 @@ Juno 0.2.3
 Compat 0.18
 MacroTools 0.3.6
 AutoHashEquals 0.1.0
+MNIST 0.0.2
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,4 @@
 julia 0.5
-Requests 0.3.10
 BinDeps 0.4.4
 ProtoBuf 0.3.0
 PyCall 1.7.1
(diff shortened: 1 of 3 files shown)


oxinabox:
Down from 38 dependencies to 26.
So managed to cut out 12.

I think this should speed up installation (and thus CI testing),
as well as make things less likely to break

REQUIRE
```diff
@@ -1,14 +1,9 @@
 julia 0.5
-Requests 0.3.10
-BinDeps 0.4.4
 ProtoBuf 0.3.0
 PyCall 1.7.1
 Distributions 0.10.2
-Conda 0.2.3
 JLD 0.6.3
 FileIO 0.1.2
-MNIST 0.0.2
```
malmaud: Can we keep this one? It's used by the MNIST examples and is just a tiny dependency.

Issue #220, opened 2017-05-23
Build TensorFlow from source


aterenin:
Hi,

For maximum performance, it is highly recommended to build TensorFlow from source. This will also be needed for Mac OS X GPU support, as Google said they will not be releasing a binary going forward.

I suggest adding such an option. I suggest using Homebrew.jl on OS X - there is already a formula for installing TensorFlow available at https://github.com/Homebrew/homebrew-core/blob/f7108950af85e222b11523c7ae67140736faa062/Formula/libtensorflow.rb that will probably need minor edits for GPU support. I am not familiar with how Julia handles dependencies on Linux, but perhaps somebody else is.

Thanks for looking.

oxinabox:
There are instructions in the docs for building tensorflow from source.
There is also a script in `/deps` to do it.
The `/deps` script could I guess be set to run automatically based on an enviroment variable.

I actually can't get that script to run,
but I am not very familiar with docker.

aterenin:
Thanks, just saw that script. I believe that is for Linux versions only. 

Building via Docker may also result in performance loss, albeit a much smaller one compared to downloaded binaries. Assuming Docker doesn't do anything unexpected, it should compile with SSE instructions, correct CUDA compute capability, etc, but may not link to Intel MKL, etc, if the user has them installed. This would make it far better than precompiled binaries, but not as good as building without Docker. On the other hand, Docker makes managing dependencies easier, so it may be the simpler solution - I am not familiar with how Julia handles this on Linux.

Since this is already implemented on Linux, perhaps we only need to consider Mac OS X. With Homebrew.jl it should be straightforward - just download the correct version of the formula and run `brew install --build-from-source libtensorflow`. It would definitely be helpful to have a environment variable for running that script automatically.

aterenin:
A quick update. Adding the following to the libtensorflow.rb formula results in a successful GPU build on my system.


    url "https://github.com/tensorflow/tensorflow/archive/v1.0.0-rc1.tar.gz"
    sha256 "55d657e5f9d3a3630e616d6928ff648172c803004541066105beef3b7d685207"

and

    ENV["TF_NEED_CUDA"] = "1"
    ENV["GCC_HOST_COMPILER_PATH"] = "/usr/bin/gcc"
    ENV["TF_CUDA_VERSION"] = "8.0"
    ENV["CUDA_TOOLKIT_PATH"] = "/usr/local/cuda/"
[...]

PR #257, opened 2017-06-15
Added documentation for building from source (with or without GPU support)

aterenin:
After the discussion in Issue #220, I figured it'd make sense to write a documentation page on building from source. Currently, the information needed for everything to work is scattered in between TensorFlow documentation pages and various issues on this repo.

It would be great to add some information about the binary patch in #245 so that users may incorporate it. That way potential segfaults and unexpected behavior are avoided.

### DIFF
--- a/docs/src/build_from_source.md
+++ b/docs/src/build_from_source.md
@@ -1,0 +1,48 @@
+# Building TensorFlow from source
+
+Building TensorFlow from source is recommended by Google for maximum performance - on some systems, the difference can be substantial. It will also be required for Mac OS X GPU support for versions later than 1.1. This document describes how to do so.
+
+
+## Step 1: build libtensorflow
+
+To build libtensorflow for TensorFlow.jl, follow the steps here: https://www.tensorflow.org/install/install_sources, except for a few minor modifications.
+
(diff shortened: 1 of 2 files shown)


docs/src/build_from_source.md
```diff
@@ -0,0 +1,48 @@
+# Building TensorFlow from source
+
+Building TensorFlow from source is recommended by Google for maximum performance - on some systems, the difference can be substantial. It will also be required for Mac OS X GPU support for versions later than 1.1. This document describes how to do so.
+
+
+## Step 1: build libtensorflow
+
+To build libtensorflow for TensorFlow.jl, follow the steps here: https://www.tensorflow.org/install/install_sources, except for a few minor modifications.
+
+  * Ignore the step "Install python dependencies".
```
oxinabox: I suggest maybe:

 Ignore the step "Install python dependencies" -- these will not be needed as we are not building  for python. (You do still need to install Bazel though.)

aterenin: Done

aterenin:
All changes should be addressed in latest commit.

oxinabox:
Thank-you.
This is great.

PR #259, opened 2017-06-16
use Conda.jl to install python binding

### LINKED ISSUE #252: Use Conda.jl to install Python TensorFlow
It should be very easy to install the Python TensorFlow using Conda.jl
Probably only adding 2 lines to `build.jl`

Is there any reason not to do this?
It would make the setup easier for new users since they won't have to worry about installing anything.

oxinabox:
Closes #252 

Also, if we are lucky this might actually even install Cuda and cudnn.
Which would be *excellent* improvement to user experience.

Lets see if this passes on Travis


oxinabox:
OK working for linux, not working for OSX.
Because no tensorflow-1.1.0 for OSX found.

malmaud:
Yay, so clean

PR #260, opened 2017-06-16
WIP: Support Python3 CI

Ismael-VC:


oxinabox:
Thanks for your help working this out.
As per our discussion: I made the PR:  https://github.com/JuliaPy/Conda.jl/pull/74
adding support for basically this approach.

Given that using the pybuildtensorflow.jl script is mutually incompatible with
#259 I think better to wait til #259 and the PR on Conda.jl or similar is merged, and 
then make the travis.yml changes for that.
So closing this now.
Thanks again for the help.

PR #274, opened 2017-07-07
WIP: Support Julia master

malmaud:


src/core.jl
```diff
@@ -1045,7 +1046,7 @@ function load_proto(tensor::tensorflow.TensorProto)
                 push!(val, val_i)
             end
         else
-            val = reinterpret(eltype(val), tensor.tensor_content)
+            val = Vector{eltype(val)}(unaligned_reinterpret(eltype(val), tensor.tensor_content))
```
oxinabox: Probably just:
`Vector(unaligned_reinterpret(eltype(val), tensor.tensor_content))`
would do?

src/meta.jl
```diff
@@ -20,10 +20,10 @@ macro op(f)
         end => opname
     end
     opname === nothing && error("Invalid usage of @op")
-    # opname = f.args[1].args[1]
+
     already_registered = opname ∈ registered_ops
     push!(registered_ops, opname)
-    @assert(isdefined(:tf)) # Need tf as name for module where this code is located
```
oxinabox: Why don't we need this anymore? Won't this let a bunch of things break?
Particularly if people start wanted to define their own ops outside this package.
(eg I think I have a `@op Base.atan2(x,y;name=nothing)=....` somewhere in my code)

malmaud: Ah ya, it raises a deprecation warning on .7 though. 

Instead of just addressing the deprecation though, can we think of a more elegant solution than requiring that user code has a certain symbol defined>

src/core.jl
malmaud: yep

src/meta.jl
oxinabox: Yes indeed.
I want to think of a more elegant way of doing this whole thing.
(Eg I dislike that we have the Holy traits, and the list, and that there are corner cases where they disagree.)

oxinabox:
@malmaud what is the status on this?

malmaud:
Superseded

PR #278, opened 2017-07-19
bugfix for top_k

yeesian:


oxinabox:
Would you care to add a one-line test to prove that it is fixed (and guard against future regressions)

PR #280, opened 2017-07-25
Check for tensorboard in Conda

### LINKED ISSUE #279: Visualize() does not work
Visualization does not work if tensorboard exists but is not in the system path. See [here](https://discourse.julialang.org/t/tensorflow-jl-how-to-use-visualize/5017/2).

malmaud:
Closes #279

### DIFF
--- a/src/show.jl
+++ b/src/show.jl
@@ -1,4 +1,5 @@
 using Juno
+using PyCall
 import Juno: Tree, Row, fade, interleave
 
 @render Juno.Inline t::Tensor begin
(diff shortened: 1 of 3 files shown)


src/show.jl
```diff
@@ -132,26 +133,51 @@ const tensorboard = Ref{Tensorboard}()
...
+
+Checks everywhere in the system path, as well in the binary python directory
+associated with the Conda.jl installation, should it exist.
+"""
+function find_tensorboard()
+    path = nothing
+    dirs = split(ENV["PATH"], ":")
+    if PyCall.conda
+        @eval import Conda
+        push!(dirs, Conda.BINDIR)
```
oxinabox: Rather than using `@eval` to bring load Conda into `main`.
Maybe just use it via where it is already imported in PyCall.

i.e. `push!(dirs, PyCall.Conda.BINDIR)`

This feels less dirty to me than `@eval`

src/show.jl
```diff
@@ -132,26 +133,51 @@ const tensorboard = Ref{Tensorboard}()
...
+"""
+    find_tensorboard()
+
+Return the path to the tensorboard executable.
+
+Checks everywhere in the system path, as well in the binary python directory
+associated with the Conda.jl installation, should it exist.
+"""
+function find_tensorboard()
+    path = nothing
```
oxinabox: Do `local path` rather than `path=nothing`
so it can be type stable?
Or use a nullable?
I guess with the small union optimisation this is less of a thing now, though


`local path` would require moving the errorcheck into this function which I think might be cleaner anyway.

src/show.jl
```diff
@@ -132,26 +133,51 @@ const tensorboard = Ref{Tensorboard}()
...
+    if isdefined(tensorboard, :x)
+        port = tensorboard[].port + 1
+    else
+        port = 6006
+    end
+    if logdir === nothing
+        logdir = mktempdir()
+    end
+    path = find_tensorboard()
+    if path === nothing
```
oxinabox: I feel like this error should maybe be part of `find_tensorboard` maybe.
As you either found it, or you did not.
Not finding it it I think is an error, rather than saying you found nothing.

src/show.jl
malmaud: 👍

src/show.jl
malmaud: I wouldn't about type stability for things not in an inner loop.

src/show.jl
malmaud: I think I'll keep it this way for now, to make it easier for the more user-facing code to choose the appropriate error message. But let's revisit later if it seems inelegant.

src/show.jl
oxinabox: fair enough

PR #281, opened 2017-07-25
Warn if Python TF version is too low

malmaud:


src/TensorFlow.jl
```diff
@@ -145,6 +145,11 @@ function load_python_process(;force_reload=false)
                 init()
             end
         end)
+        py_version = tf_version(kind=:python)
+        jl_version = tf_version(kind=:julia)
+        if py_version < jl_version
+            warn("Your Python TensorFlow version ($py_version) is below your Julia TensorFlow version ($jl_version). This can cause various errors. Please consider upgrading your Python version by running `pip install --upgrade tensorflow` from the command line and then restarting Julia.")
```
oxinabox: One can not assume tensorflow was installed via pip.
It wound't be if it was installed via Conda.jl.
Message should either just say upgrading or give conda instructions as well

src/core.jl
```diff
@@ -45,12 +45,23 @@ macro required(keywords...)
...
+Return the version number of the tensorflow library.
+
+If `kind` is `:julia`, return the version of the TensorFlow binary loaded
+into the Julia process.
+
+If `kind` is `:python`, return the version of the Python API.
 """
-function tf_version()
-    res = @tfcall(:TF_Version, Cstring, ()) |> unsafe_string
+function tf_version(; kind=:julia)
```
oxinabox: Possibly this should be 3 options:

 - `:C` for getting the shared library version (or `:backend`), i.e. current behavour of `:julia`
 - `:python` for getting python version
 - `:julia` for gettings version of TensorFlow.jl (just call `Pkg.installed("TensorFlow")`)

src/TensorFlow.jl
malmaud: 👍

src/core.jl
malmaud: 👍

PR #282, opened 2017-07-25
Check backend version against python client version

oxinabox:
I think this was missed when changing the kwarg value to `:backend`.
in #281


malmaud:
Ah woops, thanks. I wonder how we should test this.

oxinabox:
Testing this is not worth it, IMO.

The way to do it is have the test script change the backend library (and in turn the python library) using `run` commands.
And also have it able to back those changes out.

PR #283, opened 2017-07-26
Improve softmax loss docstrings

oxinabox:
Minor documentation change.
Because I forget that these things took kwargs only now

### DIFF
--- a/src/ops/nn.jl
+++ b/src/ops/nn.jl
@@ -196,7 +196,7 @@ end
 
 
 """
-`softmax_cross_entropy_with_logits(logits, labels, name=None)`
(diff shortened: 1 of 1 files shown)


src/ops/nn.jl
```diff
@@ -220,6 +220,9 @@ output of `softmax`, as it will produce incorrect results.
 `logits` and `labels` must have the same shape `[batch_size, num_classes]`
 and the same dtype (either `float32` or `float64`).
 
+**Note that to avoid confusion, it is required to pass only named arguments to
```
malmaud: Let's tweak the grammar a bit. How about something like:

This function only accepts keyword arguments to avoid the frequent confusion that positional arguments cause.

PR #284, opened 2017-07-26
Add a check that dropout has working shape inference

oxinabox:
In the course of chasing down another issue,
I wanted to be sure it wasn't dropout failing shape inference.

Indeed dropout is not failing shape inference,
and now we have tests to prove it.

### DIFF
--- a/test/shape_inference.jl
+++ b/test/shape_inference.jl
@@ -194,10 +194,18 @@ end
     end
 end
 
+
 @testset "dropout" begin
     for var in (m,n,k)
         @test get_shape(nn.dropout(var, 0.5)) == get_shape(var)
     end
+
+    v = placeholder(Float32, shape=[-1,10])
+    w = placeholder(Float32, shape=[10,5])
+    vw = nn.dropout(v, 0.5*i)*w
+
+    @test get_shape(vw) == TensorShape([-1,5])
+
 end
 
 @testset "Ensure broadcasting operations do not change shape (issue #285)" begin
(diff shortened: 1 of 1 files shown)


oxinabox:
Found it.
Dropout probably shouldn't care about the shape of the keep_prob argument.
It can probably assert that it is always a scalar.

But anyway, right now it does, and this test proves it.
And increases our coverage

malmaud:
There's a conflict

oxinabox:
Rebased

test/shape_inference.jl
```diff
@@ -194,6 +194,21 @@ end
...
+    for var in (m,n,k)
+        @test get_shape(nn.dropout(var, 0.5)) == get_shape(var)
+    end
+
+
+    m_by_i = nn.dropout(m, 0.5*i) # a scalar keep_prob
+    @test get_shape(m_by_i) == get_shape(m)
+
+    m_by_n = nn.dropout(m, 0.5*n) # a fully unknown keepprob
+    @test get_shape(m_by_n).rank_unknown == true
```
malmaud: Is this just a limitation of the current shape inference code, or the truly desired behavior?

oxinabox: A limitation of the current shape inference code.
This PR proves that that limitation is there.

I would use `@test_broken ... = false` but I think that causes Travis to fail

malmaud: I think we should take it out, or at least comment it out.  That's based on the theory that the point of tests is to ensure that new patches don't interfere with desired and documented behavior; their point isn't just to catalog the way things happen to work now.

PR #287, opened 2017-07-27
Fix #285

### LINKED ISSUE #285: Very unusual behavour regarding static loops and scalar placeholders
Ok, this is really weird.
I have narrowed it down from something complex that involved generating a deepnet with dropout.

This bug seems to occur, if one is both multiplying a matrix tensor with a scalar placeholder **and** displaying the value of that tensor, in a loop.
(Not 100% sure loop is required)

Edit: 
there is a mistake in one of the examples. Posted here.
To avoid confusion I have posted a new clearer example below

These are weird happenings,
so it is quiet likely a mistake on my part, but I can't see it.

oxinabox:
<S> I am attempting to debug #285 </S>
I have fixed #285

In doing so I have been going through a lot of shape inference related code.
and have done some refactoring

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -1108,18 +1108,10 @@ function load_proto(tensor::tensorflow.TensorProto)
 end
 
 function load_proto(shape::tensorflow.TensorShapeProto)
-    dims = Nullable{Int}[]
     if shape.unknown_rank
         ShapeInference.TensorShape(nothing)
     else
-        for dim in shape.dim
-            if dim == -1
-                push!(dims, Nullable{Int}())
-            else
-                push!(dims, Nullable(dim))
-            end
-        end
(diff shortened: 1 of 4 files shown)


src/core.jl
```diff
@@ -1108,18 +1108,10 @@ function load_proto(tensor::tensorflow.TensorProto)
...
     else
-        for dim in shape.dim
-            if dim == -1
-                push!(dims, Nullable{Int}())
-            else
-                push!(dims, Nullable(dim))
-            end
-        end
-        ShapeInference.TensorShape(dims)
+        ShapeInference.TensorShape(shape.dims)
```
oxinabox: I'm not certain this works.
If the previous code worked then this should would too.
But IDK if the previous code worked either

PR #288, opened 2017-07-28
Ensure save and restore work as per the docs

oxinabox:
Example in docs was not running for me.
For a few reasons.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ docs/build
 *.*~
 
 *.swp
+*.jld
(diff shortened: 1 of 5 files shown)


oxinabox:
Also importing graph was broken, see that commit.
I am not sure it is the right solution but it fixes it.

Issue was that both `extend_graph`, and `import_graph` want to take a list of nodes and process them one at a time.
And `import_graph` wants to give just a single node to `extend_graph`

PR #289, opened 2017-07-31
Test against Python 2 and Python 3 -- fix #251

### LINKED ISSUE #251: Setup a Python 3 test enviroment on CI
I think we should setup a continuous integration environment with Python 3.
Current TravisCI tests only Python 2

The issues that arose during the dev/test of #248 are (I believe),
the result of the dev machine using Python 3, but the CI testing machine using Python 2.

As we want to support users using Python 2 and Python 3,
we should test on both of them.

oxinabox:
Ok, now that we have a new version of Conda that supports this.
We can do this.
Time to double all our testing times.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,8 @@ julia:
   - 0.6
   - nightly
 env:
-    - PYTHON=""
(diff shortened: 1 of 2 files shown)


malmaud:
Yay

Issue #204, opened 2017-04-12
Windows support?


mbeach42:
Now that TensorFlow supports Python 3.5 on Windows, is it possible for TensorFlow.jl to be supported on Windows?

niczky12:
I'd like that too, please

malmaud:
I would say this is a low priority for me personally, but the kind of thing I'd welcome contributions on. Perhaps in a month or two I can look into it myself.

oxinabox:
It is also a low priority for me.

One thing I would like to try out,
is if this works inside the Linux Subsystem for Windows 10.
Julia does (infact I hear it works a lot better than julia does in windows generally).
and I have heard CUDA does.
If anyone tries it out, I would appreciate a report back. Can mention it in the docs.

niczky12:
Only have win7 unfortunately, so can't try this unfortunately. Would be happy to do testing otherwise.

malmaud:
It at least works in CPU-only mode on the Linux Subsystem. Haven't tried
GPU yet.

niczky12:
Can someone point me the direction of some scripts which show how I could use `PyCall` to run tensorflow from Julia? I'm having troubles replicating the usual `with graph.as_default():` block of the Python call within Julia.

malmaud:
https://github.com/malmaud/TensorFlow.jl/blob/master/src/py.jl#L19 shows how to use Python's `with` from PyCall.

HerbM:
Does anyone have any idea how difficult an issue this might be?  Anyone looked at it at all?
It's probably beyond my time and (current) knowledge but I would like to see it work also.
Windows 7 64-bit, Julia 1.0
(It's will harm the adoption of Julia if lots of issues make Windows a 2nd-class citizen on Julia -- not a criticism of the developers and maintainers of this package or any other specifically but as a group it's important to maintain mind share across all platforms.  
I can switch to Linux but not everyone can, and I don't really want to do that.)
To really look at this though I am going to need to get a dev environment tested first probably.

HerbM:
The following might not be relevant, but I started in with the TensorFlow install/update and then the test offered on the TensorFlow site the resulting warning explicitly references "segmentation faults":
 
    > The HDF5 header files used to compile this application do not match
    > the version used by the HDF5 library to which this application is linked.
    > Data corruption or **segmentation faults** may occur if the application continues.

While lots of things may cause segmentation faults, it is interesting that this is the main problem mentioned above.

More complete message:
[...]

PR #302, opened 2017-08-18
[Failing] Windows support

oxinabox:
I wanted to see exactly how far we are from having windows support. (#204)

This is roughly what it looks like.
the PR  includes a build script,
which downloads the libtensorflow from google.
(This means that until https://github.com/tensorflow/tensorflow/issues/8669 is solved upstream we won't have dynamic RNNs, because they rely on @malmaud's custom patch.)

One particular problem is that it requires Python 3.5.
Which AFAICT one can only get our of Conda.jl by doing a re `Pkg.build` of it with the environment variables `CONDA_JL_VERSION` set to 3 (rather than the default 2).

The majority of tests pass.

The tests which fail are the ones from:    "math.jl",   "meta.jl",     "control.jl",     "nn.jl", and    "training.jl".
And they don't so much fail, as segfault.

One minimal example is:
```julia
julia> TensorFlow.group(placeholder(Int32))
2017-08-18 16:27:20.004158: F .\tensorflow/core/lib/gtl/inlined_vector.h:157] Check failed: i < size() (0 vs. 0)
```

All the errors are of that form: `<timestamp>: F <path the a header file> Check failed: <assertions>`.

**NB** While this is labelled WIP, it is not actually in progress.
I don't have any immediate plans to finish this off and make windows work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,7 @@ docs/build
 
 *.swp
 *.jld
+
+deps/c_api\.h
+
+deps/LICENSE
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -56,14 +56,8 @@ end
 base = dirname(@__FILE__)
 download_dir = joinpath(base, "downloads")
 bin_dir = joinpath(base, "usr/bin")
-
-if !isdir(download_dir)
-    mkdir(download_dir)
-end
-
-if !isdir(bin_dir)
-    run(`mkdir -p $bin_dir`)
-end
+mkpath(download_dir)
+mkpath(bin_dir)
 
 
 function download_and_unpack(url)
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -15,6 +15,11 @@ if is_apple() && use_gpu
     use_gpu=false
 end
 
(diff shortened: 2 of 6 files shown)


malmaud:
Looking good, thanks for working on this. Maybe one day we can get all the tests to pass.

PallHaraldsson:
Does this just work now?

Was only https://github.com/malmaud/TensorFlow.jl/pull/301 blocking?

[I'm looking into Python 2 vs Python 3 for other reasons.. PyCall.jl and all else supports either; do you know if you can use both in the same project e.g. Python 3.5 as Google's Tensorflow requires, and then 2.7 for some unrelated library?]

oxinabox:
No.
That does not work *at* all.
It Segfaults for basic operations like creating variables.

PallHaraldsson:
"One particular problem is that it requires Python 3.5."

I just wanted to let you guys know that TF supports Python 3.7 across all platforms now. I didn't look much into the Windows problem (I don't use/need it), but possibly TF now just works (or with little help) if the uneven Python support was the problem. There's also an open issue here on supporting latest PyCall. It may or may not help.

PR #308, opened 2017-08-31
Fixes to getting operation inputs

malmaud:


src/core.jl
```diff
@@ -888,10 +888,52 @@ struct Port
     index::Int
 end
 
+function get_num_inputs(op::Operation)
+    @tfcall(:TF_OperationNumInputs, Cint, (Ptr{Void},), op.ptr) |> Int
+end
+
+struct InputOutOfRange <: Exception
```
oxinabox: Is this worth having out own exception for?
vs `Base.BoundsError`
It might be, that is a nicer error message, but it is more code to maintain

Also convention is that all exception types end with the word "Error", no?

malmaud: I do like having my custom error messages, but you are right about the naming convention.

PR #310, opened 2017-09-04
Add forget bias parameter to LSTM cell

### LINKED ISSUE #299: Make LSTM forget-bias easier to specify
It is apparently super important to set your forget bias to 1.0 or greater.
See e.g.  [Jozefowicz et al, 2015, "An Empirical Exploration of Recurrent Network Architectures"](http://proceedings.mlr.press/v37/jozefowicz15.pdf?utm_campaign=Revue%20newsletter&utm_medium=Newsletter&utm_source=revue).

[Right now, we initialize it to be 0.](https://github.com/malmaud/TensorFlow.jl/blob/4c729522d54bdbb87bc0fcc27ec870ffcc7f32a4/src/ops/rnn_cell.jl#L155)

Python's [BasicLSTMcell](https://www.tensorflow.org/api_docs/python/tf/contrib/rnn/BasicLSTMCell) takes it as a parameter, with default value 1.0.
So does it's regular [LSTMcell](https://www.tensorflow.org/api_docs/python/tf/contrib/rnn/LSTMCell).

We should do this also.
I think the change is easy to make.
I can do so if you think it is a good idea.

oxinabox:
Closes #299

### DIFF
--- a/src/ops/rnn_cell.jl
+++ b/src/ops/rnn_cell.jl
@@ -96,7 +96,10 @@ end
 
 type LSTMCell <: RNNCell
     hidden_size::Int
+    forget_bias::Float32
 end
+
+LSTMCell(hidden_size; forget_bias=1f0) = LSTMCell(hidden_size, forget_bias)
 
 output_size(cell::LSTMCell) = cell.hidden_size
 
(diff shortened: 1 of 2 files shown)


Evizero:
currently using tensorflow from this branch. thanks!

PR #312, opened 2017-09-04
[WIP] Handle difficulties relating to while_loop

oxinabox:
I'm not longer certain that https://github.com/tensorflow/tensorflow/issues/8669 is actually an upstream issue.
At least not in the form we see it most often -- the while_loop

My MWE is:

```julia
using TensorFlow

sess = Session(Graph())
i = constant(1)
end_i = TensorFlow.while_loop((i->i <= 10), i->i+1, [i])
```

Giving: 
```
Tensorflow error: Status: Input 0 ('Const') for 'while/Enter' was not previously added to ShapeRefiner.
```

But actually this can be reproduced more simply via:
```julia
sess=Session(Graph())
j=constant(31)
g=Graph()
c_o=as_default(g) do
    c = 2*j
end
```

giving: 
```
Tensorflow error: Status: Input 1 ('Const') for 'Mul' was not previously added to ShapeRefiner.
```

Because that is basically what is happening.
We are creating a new graph, 
and then referring to nodes in the old-graph in it.

The reason we can get away with supressing the errors via the [custom patch](https://github.com/malmaud/TensorFlow.jl/blob/master/deps/build_libtensorflow/gpu/upstream_patch)
is because we then go and  transfer stuff back to the original graph such that by run-time all the references we made are in fact valid.

I'm not sure how exactly to fix this.
It'll need more thought.

This PR currently contains my refactoring of the code as I dug in and tried to understand it.

### DIFF
--- a/src/ops/control_flow.jl
+++ b/src/ops/control_flow.jl
@@ -245,62 +245,78 @@ Example using shape_invariants:
                         parallel_iterations=10, back_prop=true, swap_memory=false)
     g = Graph()
     def_graph = get_def_graph()
+    # Not transfering def_graph.shapes as shape inference on placehoders is not used here.
+    # Referentially linking everything else though.
     g.op_context = def_graph.op_context
     g.name_idx = def_graph.name_idx
     g.collections = def_graph.collections
-    # variables = map(Tensor, variables)
-    local output, g_def
+
     variable_tensors = Tensor.(get_tensors(variables))
+
+    local output, g_def
     as_default(g) do
         with_op_name(name, "while") do
             with_frame(parallel_iterations, back_prop, swap_memory) do
                 context = get_def_graph().op_context.while_context[end]
+                @assert get_def_graph() === g
+
+
+                # Preloop
+                # prepare the Enter nodes, to later merge with their update values
                 merge_nodes = Tensor[]
-                merge_names = String[]
-                output = Tensor[]
-                body_input = Tensor[]
                 enter_nodes = Tensor[]
                 for var in variable_tensors
                     enter_op = Ops.enter(var, frame_name=context.context_name)
                     push!(enter_nodes, enter_op)
                     merge_op = Ops.merge([enter_op, enter_op])[1]
+                        # For now set both inputs to be merged to be the same
+                        # But later, we will reassign one of them to use the values from the body
                     push!(merge_nodes, merge_op)
                     fillin(merge_op.op)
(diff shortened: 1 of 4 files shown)


oxinabox:
Copying over my notes from Slack (there were written before the PR) but still:

> 
> I think I am getting close to understanding why while loops are broken without patching libtensorflow. 
> 
> *  the graph for the while loop is built on a separate subgraph to the actual graph
>  * that subgraph is identical except for it is is missing the tracking for shapes of placeholders (which I think is just a typo), but **more importantly** it is a different graph on the backend. A graph that is initially empty (edited)
> * the very first thing one does is add the (so called) loop variables to that new graph, creating a bunch of Enter nodes.
> * However, that fails, because those Enter nodes take the loop variables as parameters **But** said loop variable nodes are not in this new graph -- they are back on the old graph.
> * This throws the Tensorflow error: Status: Input 0 ('Const') for 'while/Enter' was not previously added to ShapeRefiner.  because one of the things adding a Node to the graph does is add it to the backend ShapeRefiner
> * With your patch, this error is suppressed
> * It then all works out fine in the end, because all the new nodes that matter are shifted back onto the main Graph, where said loop variables are indeed defined. (edited)
> --
> So the simple solution is to stop using a separate graph while building the while-loop.
> I tried that and it did not work out so well.
> Presubaly you decided to do it this way for a reason.
> 
> So the other solution is to copy the inputs over to the sub-graph prior to adding the nodes.
> But I don't really 100% have my head around manipulating the tensorflow graphs at the 1 level above protobuff layer.

src/ops/control_flow.jl
```diff
@@ -315,20 +331,28 @@ Example using shape_invariants:
                 end
                 extend_graph(def_graph, g_def.node[to_delete])
                 deleteat!(g_def.node, unique(to_delete))
-                for collection in [:Variables, :TrainableVariables]
-                    for var in get_collection(g, collection)
-                        name = get_def(var.var_node).name
-                        as_default(def_graph) do
-                            Variable(name)
-                        end
+                @assert get_collection(:TrainableVariables) ⊆  get_collection(:Variables)
```
oxinabox: This line throws an error on travis:
`TypeError: non-boolean (TensorFlow.Tensor{Bool}) used in boolean context`

Have we got an overload that breaks this, or am I making a silly mistake?

malmaud: Mmm not sure. I think I'll just merge for now and then try to fix it.

PR #325, opened 2017-09-27
Make String tensors scalar like

andreasnoack:
Was trying out the [image recognition tutorial](https://www.tensorflow.org/tutorials/image_recognition) a had some issues with the string tensor. I'm new to TensorFlow but I think these changes are necessary to get the right behavior for string tensors. At least it was necessary for the tutorial code to work and I think the intermediate results are similar to the Python version. Do you handle string tensor in a special way in this package?

### DIFF
--- a/src/run.jl
+++ b/src/run.jl
@@ -122,6 +122,9 @@ end
 
 cast_type(T, val) = val
 
+tf_shape(x::Union{Number,String}) = TensorShape(Int[])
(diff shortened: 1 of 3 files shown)


oxinabox:
Can this have a test case added?
(In particular one that was broken before this change)

andreasnoack:
I've added some tests

malmaud:
Thanks!

PR #327, opened 2017-09-28
Better license file

PallHaraldsson:


PallHaraldsson:
Yes, this seems trivial.

Not saying MIT license should be changed to Apache (I think not a requirement), but in case dependency license is stricter (e.g. for patents and GPL compatibility) then should be obvious reading the first lines somehow?

malmaud:
I'm pretty happy with how the license is now. Thanks, though.

PR #330, opened 2017-10-03
Fixes metagraph importing.

### LINKED ISSUE #306: `import_meta_graph` segfaults on MetaGraph created from Python
```python
using PyCall
@pyimport tensorflow as tf
x = tf.placeholder("float32")
y = tf.Variable(1.0)
z = x + y
tf.train[:export_meta_graph]("test.meta")

using TensorFlow
TensorFlow.train.import_meta_graph("test.meta")
```

malmaud:
Closes #306.

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -348,9 +348,10 @@ end
         if isnull(get_node_by_name(graph, node_def.name))
             # First try to directly add this node to the graph
             try
-                Operation(node_def)
+                new_op = Operation(node_def)
                 continue
             catch err
+                DEBUG_EXTEND_GRAPH && warn(err)
             end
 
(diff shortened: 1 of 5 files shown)


src/core.jl
```diff
@@ -348,9 +348,10 @@ end
         if isnull(get_node_by_name(graph, node_def.name))
             # First try to directly add this node to the graph
             try
-                Operation(node_def)
+                new_op = Operation(node_def)
```
oxinabox: `new_op` variable is never used AFAICT, so keeping it is not useful.
As an aside this function is crazy long, it would be nice to break it into several subfunctions.

malmaud: Yep, this should definitely be refactored sometime.

Issue #221, opened 2017-05-23
Make `setindex!`  work


oxinabox:
Make  give natural julia syntax for setting elements of tensors using `scatter_nd` etc

 I have been thinking about making what I have been calling in my head "The big Assignment update" for TensorFlow.jl

 This basically means ensuring that all the `scatter` and `assign` operations are imported, then setting up all the overloads for `setindex`.

Currently I think the only overload is `a[]=b` becomes `assign(a,b)`

 The main target to add is eg `a[:,3:end, 5]=b` becoming `scatter_nd_update(a,b, MATH)`, where `MATH` is the code to be written to do the index transformation

What I am not sure how to handle is `a[:,3:end, 5] += b`
 and for the `a[]+=b` there exists the `assign_add` operation
 but  `+=` is not  overloadable julia?

 **Option A)** is to ignore it. It is probably a fairly minor speed difference vs `assign` + `add`  (See https://stackoverflow.com/q/44093266/179081)

 **Option B)** is to do a julia level rewrite, via the `@tf` macro. to pickup the uses of `+=` and replace them with function calls eg to `_pluseq!` which then have the dispatchs defined on them

 **Option C)** is add a whole "graph optimization" layer to TensorFlow.jl that detects operation combinations that can be simplified like (assign, add) and does a graph level rewrite before handing the graph over to the C/C++ backend.
But this seems like implementing something that should be defined in the backend. (Like XLA?).

Cross-ref: the same for getindex: #166

RatanRSur:
Hey, was looking into this to see if I could put together a PR for this if you haven't started it. Are there any complications you forsee for the first part? The `setindex!` overloads with `assign` and `scatter` seem pretty straightforward.

As for the second part, it is an interesting problem considering julia's `+=` is syntactic sugar. Where is the `_pluseq!` you are referring to?

Also, it's been a while since I looked at julia, what does the `a[] += b` syntax mean? How does it differ from `a += b`?

oxinabox:
I have not started it,
I suggest though before making a PR on this, waiting until I refactor the `getindex` code.
Right now it is a bit messy as it has to work around some limitations WRT triangular dispatch in julia 0.5.
Once I clean it up for 0.6, then anyone working on doing `setindex!`  would have a much stronger starting point.
Since the code is similar, substuting `scatter` for `gather`, and `scatter_nd` for `gather_nd`,
as you say.

[...]

PR #332, opened 2017-10-05
Refactor GetIndex to use triangular dispatch

oxinabox:
Nothing I like as much as being able to delete a whole pile of ugly code.
This passes all tests, which I believe were pretty comprehensive,
so I believe these changes are thus all fine.

This was much easier than expected.

This is marginally blocking for #221

### DIFF
--- a/src/ops/indexing.jl
+++ b/src/ops/indexing.jl
@@ -3,7 +3,8 @@
 
 
 """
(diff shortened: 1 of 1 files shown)


oxinabox:
cf: https://github.com/JuliaLang/julia/issues/23996

malmaud:
Wow, this is really elegant.

PR #342, opened 2017-10-24
Define `Base.close(::Session)`

malmaud:


src/run.jl
```diff
@@ -77,9 +77,18 @@ function build_input(tensor_map::Dict)
     input_tensors, input_values
 end
 
+struct ClosedSessionError <: Exception
+end
+
+function Base.show(io::IO, err::ClosedSessionError)
+    print(io, "Tried to call 'run' with a closed TensorFlow session.")
```
oxinabox: This is assuming the call-stack.
Right now this can only be thrown by `run`, true.
But the user will find that out by looking at the call-stack.

I think it should be "The TensorFlow session is closed.")

malmaud: Good point

malmaud:
Closing sessions is important for freeing the GPU resources. The resources attached to the other TensorFlow objects is pretty trivial. I guess it couldn't hurt to add `close` methods to them sometime though.

PR #345, opened 2017-10-29
squaring of global_norm adaptive to data type

kevin-w-li:
I noticed there is a `.^2.0` which would always produce `Float64`. Not sure if this is the best way to deal with it but let me know otherwise.

### DIFF
--- a/src/ops/clipping.jl
+++ b/src/ops/clipping.jl
@@ -50,7 +50,7 @@ end
     local out
     tensor_value(t) = Tensor(t)
     tensor_value(t::IndexedSlices) = t.values
-    square(t) = muliply(t, t)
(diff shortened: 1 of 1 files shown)


kevin-w-li:
should've used `t.*t`

malmaud:
What was wrong with `.*` instead of `multiply`? I'd generally prefer the former unless it has a problem.

kevin-w-li:
I tried that but it seems `.*` translates into `matmul` and complained about shape mismatch... I saw line 16 used `multiply` so I just followed.

malmaud:
A commit from a few days ago should have fixed that, but this is fine
as-is. Thanks!

Issue #20, opened 2016-09-14
Remove dependence on Python


malmaud:
This relies on the C API supporting adding gradient nodes, which Google has said will happen sometime.

spencerlyon2:
Any timeline on this?

malmaud:
No. There is a just TODO in the readme for the C API that says to add support support for gradient calculations.

MikeInnes:
In the mean time, would you be amenable to splitting out, say, TensorFlowCore.jl to provide the core API and have TensorFlow.jl wrap that + provide extras like `gradients`? It'd be pretty useful for applications that don't need `gradients` not to have the hard python dependency.

Alternatively, perhaps we could do something like lazy-loading the python process when `gradients` is first called?

malmaud:
I'd be up for lazy-loading the python process. In practice I would think most uses of TensorFlow would use `gradients`, but if nothing else, you'd get a faster start-up time.

oxinabox:
I think this might be unblocked now:
https://github.com/tensorflow/tensorflow/issues/6268

malmaud:
It is is closer, but still far away - many gradients are still missing from the C API.

PR #356, opened 2017-12-12
Update ErrMsg text from TensorFlow in core.jl

rickhg12hs:
Fixes one test with current libtensorflow.so.

### DIFF
--- a/test/core.jl
+++ b/test/core.jl
@@ -129,7 +129,7 @@ end
             gradients([u, s, v], x)
         catch ee
             errmsg = repr(ee)
-            @test contains(errmsg, "No gradient defined for operation 'Svd'")
+            @test contains(errmsg, "SVD gradient is not implemented for compute_uv=True and full_matrices=False.")
         end
     end
 end


malmaud:
cc @oxinabox, who wrote that test.

rickhg12hs:
Hmmmm, it looks like Travis's `libtensorflow.so` is different.  After my edit, the test passes on my machines.

oxinabox:
we could remove the try-catch and just use a `@test_throws`
I wanted to be more specific for some reason there.
But if tensorflow is going to keep changing the error messages from version to version...

rickhg12hs:
It looks like the error/exception message came from a TensorFlow Python (???) file that was updated just a few months ago.  It seems that some SVD gradients are supported now.

https://github.com/tensorflow/tensorflow/blame/b3d5ec90bc6b7ae7822ea82d82b41e529c5e047a/tensorflow/python/ops/linalg_grad.py#L267

I'm missing the "big picture" here.  How does a Python exception message get emitted from a Julia program?

oxinabox:
> I'm missing the "big picture" here. How does a Python exception message get emitted from a Julia program?

Most of TensorFlow.jl,   is either fully written in julia (e.g shape inference, all the RNN stuff),
or backing on to code in the TensorFlow C api.

The exception to this is the `gradients` function.
That falls back on the Python API. See #20.

oxinabox:
Anyway, I think that whole testset can just be deleted.

oxinabox:
I am pretty sure this was solved elsewhere now.

PR #359, opened 2018-01-17
WIP: Bump TF version

malmaud:


malmaud:
Getting some errors about cycles of length 1, not sure what's up with that yet.

src/ops/control_flow.jl
```diff
@@ -250,6 +250,7 @@ Example using shape_invariants:
 """
 @op function while_loop(condition, body, variables; name=nothing, shape_invariants=nothing,
                         parallel_iterations=10, back_prop=true, swap_memory=false)
+    error("Dynamic while loops are not currently ipmlemented.")
     g = Graph()
```
oxinabox: Maybe comment out the body with a multiline comment (after the error),
so that synax highlighter will make it clear that that code is not being run?

test/math.jl
```diff
@@ -234,25 +234,25 @@ end
...
-    b = [0.02]
-    v = constant([1.,2.])
-    @test_throws TensorFlow.TFException run(sess, v[1] * a)
-    @test 0.01 ≈ run(sess, v[1] .* a) |> first
-    @test 0.02 ≈ run(sess, v[1].* b) |> first
-    @test 0.0002 ≈ run(sess, v[1].* a .* b) |> first
-    @test 0.0002 ≈ run(sess, (v[1].* a) .* b) |> first
-    @test 0.0002 ≈ run(sess, v[1].* (a .* b)) |> first
-end
+# @testset "broadcasting operations" begin
```
oxinabox: Why are the commented out?

test/meta.jl
```diff
@@ -73,15 +73,15 @@ end
...
-        i = constant(1)
-        loop_sum = constant(0)
-        res = @tf while i ≤ 10
-            sq = i.^2
-            [i=>i+1, loop_sum=>loop_sum+sq]
-        end
-        @test run(sess, res) == [11, 385]
-    end
-end
+# @testset "While" begin
```
oxinabox: Lets use `@test_broken` rather than comment out this test.
If it is good enough for `Base` it is good enough for us, no?

Also will prevent code-rot in the other code.

test/nn.jl
```diff
@@ -59,7 +59,8 @@ end
 end
 
 
-for (rnn_fun, post_proc_outputs) in ((nn.dynamic_rnn, identity), (nn.rnn, last))
```
oxinabox: This file also lets use `@test_broken`?
It is a bit harder as it will need a conditional  to check the `rnn_fun`

test/nn.jl
```diff
@@ -153,17 +154,17 @@ for (rnn_fun, post_proc_outputs) in ((nn.dynamic_rnn, identity), (nn.rnn, last))
...
-    run(sess, minimize_op)
-end
+# @testset "rnn gradients" begin
+#     sess = Session(Graph())
+#     cell = nn.rnn_cell.LSTMCell(10)
+#     inputs = constant(randn(Float32, 5, 32, 5))
+#     out = nn.dynamic_rnn(cell, inputs)
+#     loss = reduce_sum(out[1]).^2
+#     minimizer = train.GradientDescentOptimizer(.01)
+#     minimize_op = train.minimize(minimizer, loss)
```
oxinabox: I think we can use `@test_broken` on the line that throws an error to catch it.

Maybe
```
@test_broken (minimize_op = train.minimize(minimizer, loss)) isa Any
```

malmaud:
Good idea to use `@test_broken`. I just skipped it for the `rnn_fun` loop since it didn't seem worth the extra logic involved instead of just sticking with the comment.

PR #366, opened 2018-03-15
rnn scopes

gustafsson:
Put RNN cells in op_name scopes to make visualisations of the graph more readable.

### DIFF
--- a/src/ops/nn.jl
+++ b/src/ops/nn.jl
@@ -67,6 +67,7 @@ function rnn(cell, inputs::Vector, sequence_length=nothing; initial_state=nothin
     outputs = tf.Tensor[]
     state = initial_state
     for (time_step, input) in enumerate(inputs)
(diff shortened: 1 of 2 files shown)


oxinabox:
This is a good idea.
Infact, I think I tried to do this a while ago.
But I gave up because it was breaking the tests (and I had bigger fish to fry)

This is breaking the tests,  see https://travis-ci.org/malmaud/TensorFlow.jl/jobs/353732737#L1133
If you can work out why and fix that, this would be great to have

gustafsson:
Seems when I use the variable `new_state` and `new_output` in two nested scopes they are no longer brought into the scope of the for loop so I need https://github.com/malmaud/TensorFlow.jl/pull/366/commits/d6644f023eff5e8de81581039fb9f2835034ba26

oxinabox:
LGTM, @malmaud?

PR #368, opened 2018-03-15
register_shape ArgMax ArgMin

gustafsson:
shape inference for ArgMax and ArgMin

### DIFF
--- a/src/shape_inference.jl
+++ b/src/shape_inference.jl
@@ -383,11 +383,8 @@ for func in ["ArgMax", "ArgMin"]
     register_shape(func) do op
         value_shape = copy(_get_shape(get_input(op, 1)))
         reduction_dims = get_input(op, 2)
-        if value_shape.rank_unknown
-            return [TensorShape(nothing)]
-        end
         reduction_dim_values = load_const(reduction_dims)
-        if isnull(reduction_dim_values)
+        if value_shape.rank_unknown || isnull(reduction_dim_values)
             return [TensorShape(nothing)]
         else
             dims = get(reduction_dim_values)+1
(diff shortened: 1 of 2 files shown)


src/shape_inference.jl
```diff
@@ -379,6 +379,27 @@ for func in ["Sum", "Prod", "Min", "Max", "All", "Any", "Mean"]
...
 
+for func in ["ArgMax", "ArgMin"]
+    register_shape(func) do op
+        value_shape = copy(_get_shape(get_input(op, 1)))
+        reduction_dims = get_input(op, 2)
+        if value_shape.rank_unknown
+            return [TensorShape(nothing)]
+        end
+        reduction_dim_values = load_const(reduction_dims)
+        if isnull(reduction_dim_values)
```
oxinabox: I think cleaner if you worte:

```
if value_shape.rank_unknown || isnull(reduction_dim_values)
       return [TensorShape(nothing)]
end
```

and removed the earlier if

src/shape_inference.jl
```diff
@@ -379,6 +379,27 @@ for func in ["Sum", "Prod", "Min", "Max", "All", "Any", "Mean"]
     end
 end
 
+for func in ["ArgMax", "ArgMin"]
+    register_shape(func) do op
```
oxinabox: This could be written 
`register_shape.( ["ArgMax", "ArgMin"]) do`

rather than a for loop.
But that might be being too clever.

src/shape_inference.jl
gustafsson: agreed

src/shape_inference.jl
gustafsson: I like that. Since I copied most of this block from the block above I'll introduce the same change there as well.

oxinabox: I agree

PR #369, opened 2018-03-15
register_shape Select

gustafsson:
shape inference for Select


oxinabox:
This code is correct, nice work.
(I had to relook up the definition of Select to remind myself what it is.)
Please add a test to confirm it is functional.

gustafsson:
Thanks, I added two tests.

oxinabox:
LGTM

PR #370, opened 2018-03-15
saver.max_to_keep needs to sort versions before discarding old versions

gustafsson:
I think this PR follows the original intent of the function and that `sort!` call was accidentally omitted.

Example of what this commit changes. Let's say we're saving checkpoints named

```
1
2
3
4
5
6
7
8
9
10
11
```

With `max_to_keep=5` we would expect only 7,8,9,10,11 to remain. But 10 and 11 are lexicographically sorted before 2 so (at least on OS X) `readdir` gives the following list:

```
1
10
11
2
3
4
5
6
7
8
9
```

And we'll only keep: 5, 6, 7, 8, 9 and we'll discard the last checkpoints that was saved most recently.

The solution is to not rely on the order from `readdir` and call `sort!` on the versioned checkpoints.

### DIFF
--- a/src/train.jl
+++ b/src/train.jl
@@ -239,6 +239,7 @@ function FileIO.save(saver::Saver, session::Session, path; global_step=nothing)
     end
     if length(versions) > saver.max_to_keep
         to_delete = length(versions) - saver.max_to_keep
+        sort!(versions)
         for i in 1:to_delete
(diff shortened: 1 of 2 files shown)


oxinabox:
Looks good, and sensible.
Do you want to turn your example into a testcase so we can prove it is fixed?

gustafsson:
Sure, I couldn't find a test of the `global_step` argument so I added a new one.

PR #371, opened 2018-03-15
named Variable with @tf

gustafsson:
I expect to be able to name a Variable with `@tf` just like I can name operations.

This makes
```
@tf W = Variable(data)
```
effectively translate to
```
W = Variable(data; name="W")
```

### DIFF
--- a/src/variable.jl
+++ b/src/variable.jl
@@ -77,6 +77,9 @@ function Variable(initial_value; name="", trainable=true, literal_name=false)
     return self
 end
 
+"""Variable is not an @op, but it supports the name argument"""
+tf.withname(::Type{Variable}, name) = tf.withname(tf.RegisteredOp(), Variable, name)
+
 @tf.with_def_graph function Variable(graph::tf.Graph, s::AbstractString)
     var_node = tf.Tensor(get(tf.get_node_by_name(graph, s)))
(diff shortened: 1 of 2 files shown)


oxinabox:
Can you add a test?

iirc `get_variable` does support `@tf` blocks, 
and also creates variables (I beleive in python the advice is to always use `get_variable` rather that `Variable` (because they are different), but I think (and I could be wrong) in TensorFlow.jl they are closer to the same.)

gustafsson:
right, `Variable(;name=nn)` doesn't add variable scopes to `nn` whereas `get_variable(;name=nn)` does. I think that names created with @tf should respect the variable scope, and I see that Variable is a more low-level construct that doesn't care about the scope.

So I'll close this PR and don't think it should be merged.

I do like that I can initialise the data explicitly with Variable, whereas get_variable always uses rand from some distribution. That's why I ended up using Variable instead of get_variable. But that's for another discussion.

malmaud:
You can pass a constant initializer to get_variable.

gustafsson:
Isn't `ConstantInitializer` passed to variable_scope, and requires all elements to be the same? My use case was that I already had a specific matrix (with not all elements equal) that I wanted to assign to the variable.

PR #372, opened 2018-03-15
tf.select on lstm state tuple

gustafsson:
to support dynamic length rnn

### DIFF
--- a/src/ops/rnn_cell.jl
+++ b/src/ops/rnn_cell.jl
@@ -13,7 +13,7 @@ IdentityRNNCell
 using Compat
 import ..nn
 import .nn: TensorFlow
-import .TensorFlow: Operation, get_shape, get_variable, tanh, Tensor, nn, concat, expand_dims, with_op_name, AbstractTensor
+import .TensorFlow: Operation, get_shape, get_variable, select, tanh, Tensor, nn, concat, expand_dims, with_op_name, AbstractTensor
 import .nn: sigmoid
 const tf = TensorFlow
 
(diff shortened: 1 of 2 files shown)


oxinabox:
Can you explain a bit more the motivation of this change?
The first file changed is a matter of style. (idk which is clearer).
The second defines an operation but its not being used for anything in this PR.


More broadly welcome and hello, thank you for your contributions.
I've just taken a quick look at each of them.
Come say high in the #machine-learning channel on the JuliaLang slack: http://slackinvite.julialang.org/

gustafsson:
The change about `new_state` and `new_output` should have been in #366. I've reverted that commit. Also see https://github.com/malmaud/TensorFlow.jl/pull/366#issuecomment-373688170

gustafsson:
If sequences have varying lengths given by the `sequence_length` argument then the `rnn` function uses `select` to skip updating states after the end of the input sequence. But `select` works on tensors and the state of an LSTM cell is a tuple of two tensors.

This PR makes it possible to select one of two LSTM state tuples based on some condition.

gustafsson:
Thanks for the invite, I'll pop by on slack to say hi!

oxinabox:
So are you saying dynamic length was broken for LSTM?

gustafsson:
yes

oxinabox:
Ahh now I understand what this PR is for.
 I think that would be my fault -- I use GRU more than LSTM.
 I will review this later.

gustafsson:
I should have mentioned that I was getting an error on this row for LSTM cells: https://github.com/malmaud/TensorFlow.jl/blob/fcf8300257a6b96940fd1d04985ea6d9d2fce0b0/src/ops/nn.jl#L77

gustafsson:
Np, thank you!

PR #376, opened 2018-04-02
WIP: Just enough of 0.7 to load

Keno:
DO NOT MERGE THIS BRANCH.
I'm just putting it up as a starting point for somebody who
wants to finish it.

Many things are commented out on this branch because dependencies don't work on 0.7 yet.


malmaud:
Thanks, looks like a great start. I guess we really need PyCall to work on .7 before we can advance this much further.

PallHaraldsson:
FYI: PyCall added 0.7 support "a month ago", so maybe this PR already just works?

PR #382, opened 2018-04-25
Use Hiccup.span

gustafsson:
Juno.span referred to Hiccup.span. But Hiccup is no longer used by Juno since https://github.com/JunoLab/Juno.jl/commit/5f3bd718c051adeaf8944a0782f55470b7788024.

I don't know how to add a unit test for this functionality since the results is supposed to be rendered in Juno.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -12,3 +12,4 @@ Compat 0.18
 MacroTools 0.3.6
 AutoHashEquals 0.1.0
 MNIST 0.0.2
+Hiccup 0.1.0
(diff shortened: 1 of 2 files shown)


oxinabox:
@MikeInnes   I believe you wrote the special juno display stuff, so you are best qualified to review this, if you could.

MikeInnes:
LGTM. If you want to avoid Hiccup you could also just use `Base.HTML` with a string.

oxinabox:
I'ld rather not add anything additional dependencies,
certainly not for just one line.
So lets do this the `Base.HTML` way.

malmaud:
I agree with @oxinabox. Care to make that change, @gustafsson ? Or  I can, no problem.

gustafsson:
Thanks @malmaud

PR #384, opened 2018-05-10
Remove extra sigmoid from GRU

oxinabox:
It was just wrong AFAICT
Sigmoid inside a tanh is basically the same as replacing the tanh with a sigmoid.
So it may actually have had meaningful effect.
Though I don't think so, I believe from the GRU design that that variable (which I think of as potential new state), really can have any choice of activation functions.
Unlike the gates which are constrained by the design.

Still the traditional activation function is `tanh`  there.


malmaud:
Yep, that was definitely just an accident. Good catch.

Issue #381, opened 2018-04-25
visualize() error


malmaud:
Reported by Alan Edelman:

```
visualize()
Traceback (most recent call last):
  File "/Users/alanedelman/.julia/v0.6/Conda/deps/usr/bin/tensorboard", line 4, in <module>
    import tensorflow.tensorboard.tensorboard
ImportError: No module named tensorboard.tensorboard
Traceback (most recent call last):
  File "/Users/alanedelman/.julia/v0.6/Conda/deps/usr/bin/tensorboard", line 4, in <module>
    import tensorflow.tensorboard.tensorboard
ImportError: No module named tensorboard.tensorboard
```

oxinabox:
I believe this is because we don't automatically install tensorboard via Conda.
Maybe we should.

oxinabox:
Fixed

Issue #374, opened 2018-03-20
tf.reset_default_graph()  in Julia ?


Julien281:
Hello

I'm using TensorFlow with Julia and I would like to reduce an expression with the following:

```
cross_entropy = nn.sparse_softmax_cross_entropy_with_logits( logits, labels) 

optimizer = train.GradientDescentOptimizer(learning_rate)

train_op = train.minimize(optimizer,reduce_mean(cross_entropy))
```

I have the following error:

`ERROR: LoadError: Tensorflow error: Status: Node name 'gradients/reduce_19_grad/Reshape' already exists in the Graph`

In python, there is this instruction to avoid the error: `tf.reset_default_graph() `

I don't find the similar command in Julia... Do you know it ?

Thank you

oxinabox:
Replacing your session with a new session witha new graph will do it 
`sess = Session(Graph())`
Or just replacing your graph using `set_default_graph(Graph())`
(but what is the point in a Graph without a session attached?)

We could probably trivially implement `tf.reset_default_graph()`, I'ld need to look at the docs for its exact behavour

Julien281:
There is the same error with the `Session(Graph())` but `set_default_graph()` is not defined... How defines it ?

```
ERROR: LoadError: Tensorflow error: Status: Node name 'gradients/reduce_7_grad/Reshape' already exists in the Graph

ERROR: LoadError: UndefVarError: set_default_graph not defined
```

oxinabox:
`sess = Session(Graph())`  should be called before declaring the compoents of the new graph.
Including before the placeholders

Julien281:
Thank you, the placeholders are fed correctly, but I have still my first error:

`ERROR: LoadError: Tensorflow error: Status: Node name 'gradients/reduce_grad/Reshape' already exists in the Graph`

If I add `set_default_graph(Graph())`, it is not defined:

`ERROR: LoadError: UndefVarError: set_default_graph not defined`

Maybe, there is a library to import ? Or something else ?

oxinabox:
yes there is a library to import, but it doesn't matter because that won't fix your problem, there is something else going on.
In anycase fixing your problem is not in scope for these issues.
I suggest taking your question to stackoverflow, and posting a full MWE. (Feel free to link it here, and I'll maybe take a look.)

We can leave this issue open until we decide if we want to make `tf.reset_default_graph()` in TensorFlow.jl
[...]

PR #386, opened 2018-05-11
TensorFlow 1.7

### LINKED ISSUE #385: Make TensorFlow 1.7 work
We are using 1.4
1.6 to 1.8 is latest on Conda-Forge.
1.8 was only release recently, so 1.7 is safer

Getting 1.7 will fix #381
and #374

.

oxinabox:
This will close #385 

It has the solution to the main thing I was fighting with there in it.

It is currently screwing up with 
```
  Test threw an exception of type RemoteException
  Expression: [68.0f0] == run(sess, gradients([7A, 3A], A, [constant([5.0f0]), constant([11.0f0])]))
  On worker 2:
  Python error: PyObject ValueError('Passed 1 grad_ys for 2 ys',)
  py_gradients at /home/wheel/oxinabox/.julia/v0.6/TensorFlow/src/py.jl:45
```

Which is related to this like of python:

https://github.com/tensorflow/tensorflow/blob/v1.7.0/tensorflow/python/ops/gradients_impl.py#L227-L228

which may be being caused by 
https://github.com/tensorflow/tensorflow/blob/v1.7.0/tensorflow/python/ops/gradients_impl.py#L207-L208

Which would suggest something is not working write with our PyCall, passing in objects.

That the `py_node_name`s are not coming in a python `list` or `tuple`.

Other thing this needs is the Graphy we  import, the static PB that is in the tests directory is outdated.
@malmaud  do you have the script that generated the original? I assume it was python.
If so It might be work checking the python script into the repo, and maybe even using python over commadline to generate that test file during `Pkg.test`?


oxinabox:
All working.
I just tested and this does indeed solve  #381

I am pretty sure it will solve #374 also, but it is hard as I'm not sure we have a full MWE for that. 
(I am going to close it when we merge this)

oxinabox:
Coverage changes on 

 - src/ops/rnn_cell.jl
 - src/ops/nn.jl
 - src/train.jl
 - src/shape_inference.jl

are  probably not  real and are related to some earlier PR, and CodeCov is getting confused, maybe because it doesn't have the right coverage report saved somewhere.

0% coverage on py.jl is as always a lie its just that it runs in separate process.

As an aside, I just tried switching to 1.8 and it immediately segfaults,
but so does python. So perhaps something is wrong with the version of tensorflow 1.8 on conda forge; at least wrt compat with my machine

src/core.jl
```diff
@@ -1594,13 +1586,12 @@ function gradients(y, x::AbstractArray, grad_y=nothing)
...
-gradients(y, x, grad_y=nothing) = gradients(y, [x], grad_y)[1]
-
-function c_gradients(y, x, dx=nothing)
+c_gradients(y, x, dx=nothing) = c_gradients([y], x, dx)
+function c_gradients(y::AbstractArray, x::AbstractArray, dx=nothing)
     status = Status()
     graph = get_def_graph()
     grads = Array{Port}(length(x))
-    if dx === nothing
+    if dx isa AbstractArray{Void} || dx isa Void
```
malmaud: Should we use `<:` instead of `isa` here?

src/core.jl
```diff
@@ -1571,10 +1559,14 @@ A user can provide their own initial `grad_ys` to compute the derivatives using
 
 see: [Python Docs](https://www.tensorflow.org/versions/master/api_docs/python/tf/gradients)
 """
-function gradients(y, x::AbstractArray, grad_y=nothing)
+gradients(y, x::AbstractArray, dx=nothing) = pyg_gradients(y, x, dx)
+gradients(y, x, grad_y=nothing) = gradients(y, [x], grad_y)[1]
+
+function pyg_gradients(y, x::AbstractArray, grad_y=nothing)
```
malmaud: What does the 'g' in `pyg` stand for? If it's `gradient`, then it seems redundant with the `_gradients` part of the name.

src/core.jl
oxinabox: I checked this with `@code_llvm` (well a MWE).
Either way it is removed during specialization.
(as is `==nothing`)

I think `isa` is cleanest

src/core.jl
oxinabox: 'g' stands for Not **g**etting a  clash with `py_gradients` defined in `py.jl`
I think a better name for this 
would be `py_add_gradients` and `c_add_gradients` maybe?

I messed with the names because I was trying to get `c_gradients` to work,  and wanted different overloads
but really that part of the PR could just be backed out and removed if you'ld prefer

test/core.jl
```diff
@@ -80,22 +79,19 @@ end
...
 @testset "Disconnected gradients" begin
     let
         as_default(Graph()) do
             unused = get_variable("unused", [], Float64)
             used = get_variable("used", [], Float64)
             loss = used.^2
             grad = gradients(loss, unused)
-            @test grad === nothing
-            # optimizer = train.minimize(ætrain.AdamOptimizer(), loss)
+            @test true
```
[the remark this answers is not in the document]
oxinabox: I was thinking that the actual test is not that `grad === nothing`,
it was that ` gradients(loss, unused)` didn't throw an error.
But actually, it is both, so this change should be undone.

PR #389, opened 2018-05-17
Use HTML instead of Juno.span

malmaud:


REQUIRE
```diff
@@ -12,3 +12,4 @@ Compat 0.18
 MacroTools 0.3.6
 AutoHashEquals 0.1.0
 MNIST 0.0.2
+Hiccup 0.1.0
```
gustafsson: Is this row still needed?

malmaud: Ah good catch.

oxinabox:
Looks right to me, But I having checked it in Juno.
Have you @malmaud  ? If not I can do so later

malmaud:
Get a chance to test yet?

oxinabox:
Not, yet. I only have Juno (or local Julia at all), on one computer, and some else in the lab has been using all its RAM, so I've not been on it

oxinabox:
Tested it, a space was missing, I am not 100% sure this fixes it but it should do

Issue #395, opened 2018-05-30
Python TensorFlow package


Julien281:
Hello

I updated the TensorFlow package and I have a problem...

When I run the [logistic regression example](https://github.com/malmaud/TensorFlow.jl#logistic-regression-example), there is the following error when the` train.minimize` is running:

```
ERROR: On worker 2:
The Python TensorFlow package could not be imported. You must install Python TensorFlow before using this package.
```
Following that, I found [this issue](https://github.com/FluxML/Flux.jl/issues/13) ; I ran `pip install TensorFlow` and it sounds fine:

```
sudo -H pip install TensorFlow

Requirement already satisfied: TensorFlow in /usr/local/lib/python3.5/dist-packages (1.7.0)
Requirement already satisfied: absl-py>=0.1.6 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (0.1.13)
Requirement already satisfied: wheel>=0.26 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (0.31.0)
Requirement already satisfied: astor>=0.6.0 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (0.6.2)
Requirement already satisfied: six>=1.10.0 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (1.11.0)
Requirement already satisfied: tensorboard<1.8.0,>=1.7.0 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (1.7.0)
Requirement already satisfied: gast>=0.2.0 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (0.2.0)
Requirement already satisfied: grpcio>=1.8.6 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (1.10.1)
Requirement already satisfied: termcolor>=1.1.0 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (1.1.0)
Requirement already satisfied: numpy>=1.13.3 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (1.14.2)
Requirement already satisfied: protobuf>=3.4.0 in /usr/local/lib/python3.5/dist-packages (from TensorFlow) (3.5.2.post1)
Requirement already satisfied: html5lib==0.9999999 in /usr/local/lib/python3.5/dist-packages (from tensorboard<1.8.0,>=1.7.0->TensorFlow) (0.9999999)
[...]

PR #396, opened 2018-05-30
conda-forge doesn't have tensorflow=1.7.0, but it does have 1.8.0

gustafsson:
Getting the following error message during `Pkg.build("TensorFlow")` on `v0.9.0`
```
Warning: 'conda-forge' already in 'channels' list, moving to the top
Solving environment: failed

PackagesNotFoundError: The following packages are not available from current channels:

  - tensorflow=1.7.0
```

I can't find 1.7.0 on [anaconda.org/conda-forge](https://anaconda.org/conda-forge/tensorflow/files) either. But it does have `tensorflow=1.8.0`.

However the 1.8.0 package doesn't seem to install `numpy` as it should. I'm getting this message:
```
julia> pyimport("tensorflow")
...
pyimport sais ImportError('No module named numpy',)
...
```
Installing numpy explicitly seems to do the trick.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -39,9 +39,8 @@ end
 
 if PyCall.conda
     Conda.add_channel("conda-forge")
-    Conda.add("tensorflow=" * cur_py_version)
     Conda.add("tensorboard=" * cur_py_version)
-    Conda.add("numpy") # unclear why this is needed
+    Conda.add("tensorflow=" * cur_py_version)
 else
     try
         pyimport("tensorflow")
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -39,9 +39,9 @@ end
 
 if PyCall.conda
     Conda.add_channel("conda-forge")
-    Conda.add("numpy") # missing dependency of the conda tensorflow package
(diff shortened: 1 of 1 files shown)


oxinabox:
I feel like something funky is going on with Conda.
Note this build is failing on CI as per https://github.com/malmaud/TensorFlow.jl/issues/395

And Conda Forge definitely used to have 1.7.0
I bumped to 1.7.0 because 1.8.0 was segfaulting for me

oxinabox:
It might be that before I was hitting https://anaconda.org/anaconda/tensorflow/files
for 1.7

By default that should be on the channel's list already.
it is odd that it wouldn't install for you.

oxinabox:
In anycase, I'ld love to have 1.8.0 working,"
but I think it is going to be a bit more than just uping the version numbers.
If not, then not and that is awesome.

So if you can get this to pass on CI, I'ld be pretty happy

gustafsson:
I have 1.8.0 passing the unit tests locally. Both with the binary from conda as well as with 1.8.0 built from source. I'll see if I can get it passed CI.

oxinabox:
Excellent.

PR #405, opened 2018-06-21
Add a versionoinfo function that dumps out all the info needed for debugging

oxinabox:
This is to help with making it easier to process issues.

Basically the user can run `tf_versioninfo()` and it outputs basically all the info that might possibly be relevant.
Some of the info will be irrelevant, like checking the `pip --version` if it was installed with Conda.
But I figured it is easier just ignore the irrelevant parts.
Alternatively, it could be made slightly more complex and avoid including things that probably don't matter.
But it is kind of hard to tell what interesting setups people might be using.

Example output:

```
julia> tf_versioninfo()
Please copy-paste all this info into any bug reports.
Note that this may display some errors. But not all errors are bad. Some are expected

----------------
Library Versions
----------------
tf_version(kind=:backend) = v"1.7.0"
tf_version(kind=:python) = v"1.7.0"
tf_version(kind=:julia) = v"0.9.0"

---------------------
Environment Variables
---------------------
Trying to evalute ENV["TF_USE_GPU"] but got error: KeyError("TF_USE_GPU")
Trying to evalute ENV["LIBTENSORFLOW"] but got error: KeyError("LIBTENSORFLOW
")

-------------
Python Status
-------------
PyCall.conda = true
PyCall.PYTHONHOME = "/home/wheel/oxinabox/.julia/v0.6/Conda/deps/usr:/home/wheel/oxinabox/.julia/v0.6/Conda/deps/usr"
readstring(@cmd("pip --version")) = "pip 10.0.1 from /usr/local/lib/python3.5/dist-packages/pip (python 3.5)\n"
readstring(@cmd("pip3 --version")) = "pip 10.0.1 from /usr/local/lib/python3.5/dist-packages/pip (python 3.5)\n"

-------------
Julia Status
-------------
Julia Version 0.6.3
Commit d55cadc350 (2018-05-28 20:20 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Xeon(R) CPU           E5520  @ 2.27GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Nehalem)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, nehalem)
```

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,4 +1,5 @@
-If making bug report please copy and paste in to the issue the output of 
+If you are making a bug report, please copy and paste the output of the following Julia snippit into the issue:
+
 ```
 using TensorFlow
 tf_versioninfo()
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,5 @@
+If making bug report please copy and paste in to the issue the output of 
+```
+using TensorFlow
+tf_versioninfo()
+```
--- a/src/version.jl
+++ b/src/version.jl
@@ -1,0 +1,109 @@
+"""
+    tf_version(kind=:backend)
+
+Return the version number of the tensorflow library.
+
+If `kind` is `:backend`, return the version of the TensorFlow binary loaded
+into the Julia process.
+
+If `kind` is `:julia`, return the version of the Julia TensorFlow package.
+
+If `kind` is `:python`, return the version of the Python API.
+"""
+function tf_version(; kind=:backend)
+    if kind == :backend
+        res = @tfcall(:TF_Version, Cstring, ()) |> unsafe_string
+    elseif kind == :python
+        res = fetch(@py_proc py_tf[][:VERSION])
+    elseif kind == :julia
+        return Pkg.installed("TensorFlow")
+    else
+        error("Kind '$kind' not recognized")
+    end
+    # Deal with version strings like "0.12.head"
+    res = replace(res, r"\.head$", "")
+    VersionNumber(res)
+end
+
+function version_check(v)
+    if tf_version() < v
+        error("You have TensorFlow binary version $(tf_version()), but need version $v to use this functionality. Please upgrade with `Pkg.build(\"TensorFlow\").")
+    end
+end
+
+"""
+    py_version_check(;print_warning=true)
+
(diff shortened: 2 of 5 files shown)


oxinabox:
Ah right version.jl needs to be after @tfcall is defined. Things just need reordering

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+If making bug report please copy and paste in to the issue the output of
```
malmaud: Grammar:
If you are making a bug report, please copy and paste the output of the following Julia script into the issue:

test/runtests.jl
```diff
@@ -40,6 +40,9 @@ if !isdefined(Base.Test, Symbol("@test_nowarn"))
     end
 end
 
+
+tf_versioninfo()
```
malmaud: Can you add a comment that you're just testing it doesn't throw an error?

src/version.jl
```diff
@@ -0,0 +1,106 @@
...
+                    " but got error: ",
+                    err
+                    )
+        end
+        nothing
+    end
+end
+
+function tf_versioninfo()
+    println("Please copy-paste all this info into any bug reports.")
```
malmaud: Wording: Please copy-paste the entirely of the below output into any bug reports.

src/version.jl
```diff
@@ -0,0 +1,106 @@
...
+                    $(Meta.quot(ex)),
+                    " but got error: ",
+                    err
+                    )
+        end
+        nothing
+    end
+end
+
+function tf_versioninfo()
```
malmaud: I'm wondering if this should be called 'debuginfo()' instead.

oxinabox: Base calls similar info for purposes of debugging julia `versioninfo()`
I wanted to call this `Base.versioninfo(TensorFlow)` but you can't dispatch on modules.

upstream tensorflow calls this `tools/tf_env_collect.sh`
we could call it something like that.
Though perhaps not aas they also display some cuda info that we don't,

test/runtests.jl
oxinabox: Not just that it doesn't throw an error, it is also useful for debugging travis failures I think,
but sure,

PR #408, opened 2018-06-29
[???] Upgrade  to 0.7

oxinabox:
This is me seeing how far we get with Femtocleaner alone


malmaud:
Ok, I think I've enabled it.

PR #412, opened 2018-07-05
[WIP] Test and update all the examples

oxinabox:
Sometimes people complain that the examples are broken.
They are correct.
We should probably be automatically testing those.

They are still broken in this PR, but now they are tested.
Don't merge til they are not broken anymore


malmaud:
Ya let's fix them!

oxinabox:
Given this has a conflict, and the earlier PRs,
I am guessing someone already fixed all of them

PR #413, opened 2018-07-06
Remove Distributions dependence

malmaud:


oxinabox:
Are you happy with the name `DefaultInitializer`
I kinda think we might as well go full in and say `RandomNormalInitializer` 
and maybe even give it a variance field (default to 0.01)

malmaud:
Ya I suppose. My thinking was I didn't want to duplicate stuff already in Distributions, but this would be so minor.

oxinabox:
It is probably worth mentioning in the docs / docstrings that any continuous distribution from Distributions.jl is a viable initializer

oxinabox:
I just rebased this because JLD.jl was borking CI,

oxinabox:
Moved Distributions and MNIST into test dependencies.
In julia 0.7 someone who installed TensorFlow.jl 
would not have access to MNIST from there environment anyway, unless they installed it separately.

malmaud:
I'm restarting the failed job

Issue #407, opened 2018-06-25
Cannot use TensorFlow as it says Failed to precompile HDF5


ebbythomas:
Hi, I am finding trouble with using TensorFlow as I  keep on getting an error "Failed to precompile HDF5"..
Any suggestions..?

Ebby

oxinabox:
I suggest attempting to `Pkg.add("HDF5"); Pkg.build("HDF5"); Pkg.test("HDF5")` 
That is the best place to start debugging.
You might need to open an issue at https://github.com/JuliaIO/HDF5.jl/

We depend on HDF5.jl (via JLD.jl) for saving and loading.
At some point in the future we might switch to JLD2.jl, or some other format, which wouldn't have the binary dependency HDF5.jl has, which might make it cause less problems for people.

ebbythomas:
Thanks,
I can see that Pkg.test("HDF5") had test errors.
Now trying to work on that
Ebby

PR #414, opened 2018-07-06
Switch to JLD2.

malmaud:


docs/src/saving.md
```diff
@@ -18,7 +18,7 @@ Then call the `save` method to save your variables:
 train.save(saver, session, "/path/to/variable_file")
 ```
 
-The newly-created "variable_file" is a [JLD](https://github.com/JuliaIO/JLD.jl) file that contains a mapping from variable names to their values. The value of variables can later be restored in a new Julia session with
+The newly-created "variable_file" is a [JLD2](https://github.com/simonster/JLD2.jl) file that contains a mapping from variable names to their values. The value of variables can later be restored in a new Julia session with
```
oxinabox: The variable file is still a JLD file.
It is just being processed and generated with JLD2.

oxinabox:
This will be less  breaky avoiding #407 and similar

docs/src/saving.md
malmaud: Oh, can JLD2 load a file saved with JLD1?

oxinabox: That was my understanding yes.

oxinabox: Ah apparently not

oxinabox:
We should really have a test for this.

docs/src/saving.md
oxinabox: Ignore me

oxinabox:
Do we want to switch to BSON.jl instead?

malmaud:
Oh ya, why not? It will probably be maintained better going forward, if nothing else. I did a quick check and they same to use the same amount of space for storing our files.

oxinabox:
If this passes, lets merge it.
I just poked the CI.

PR #424, opened 2018-08-29
Use C API for shape inference

malmaud:


src/core.jl
```diff
@@ -226,9 +211,12 @@ function with_def_graph(ex)
     error("Improper use of with_def_graph")
     (kwargs === nothing) && (kwargs = [])
     new_args = args[2:end]
+    # Temporarily eliminate passing through keyword arguments since
+    # that breaks Revise at the moment. Should add back once Revise
+    # is patched.
```
oxinabox: Can an Issue be openned to remind to fix this?
(I just stopped using Revise)

src/core.jl
```diff
@@ -1282,7 +1273,7 @@ function setindex!(desc::NodeDescription, value::TensorShape, attr_name)
     if value.rank_unknown
         dims = Int[]
     else
-        dims = Int[(isnull(dim) ? -1 : get(dim)) for dim in value.dims]
+        dims = Int[ismissing(dim) ? -1 : dim for dim in value.dims]
```
oxinabox: `dims = coalesce.(value.dims, Ref{-1})`

src/core.jl
```diff
@@ -1750,3 +1742,60 @@ function Operation(node_def::tensorflow.NodeDef)
 
     Operation(desc)
 end
+
+function set_tensor_shape(tensor::AbstractTensor, dims)
+    output = Port(tensor)
+    g = get_graph(tensor)
+    status = Status()
+    dims = [ismissing(dim) ? -1 : dim for dim in dims]
```
oxinabox: `dims = coalesce.(dims, Ref{-1})`

src/core.jl
```diff
@@ -1750,3 +1742,60 @@ function Operation(node_def::tensorflow.NodeDef)
...
+    end
+end
+
+function get_shape(tensor::AbstractTensor, dim)
+    shape = get_shape(tensor)
+    if shape.rank_unknown
+        error("Rank is unknown")
+    end
+    if length(shape.dims) < dim
+        error("$dim is more than shape's rank of $(length(shape.dims))")
```
oxinabox: This should be an `BoundsError`

src/core.jl
malmaud: I opened an issue on Revise and it was fixed there; I've deleted this comment on master.

src/core.jl
malmaud: Oh, didn't know about `coalesce`. Awesome.

PR #427, opened 2018-08-29
Add basic API for distributed session support

Keno:
I had this lying around locally for a bit. Now that julia 1.0 is supported, it's probably time to include this upstream. Note that this requires grpc support which isn't built into the tensorflow binaries that get installed by this package, so we may have to look at that separately.

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -58,11 +58,14 @@ mutable struct Status
         this
     end
 end
+Base.unsafe_convert(::Type{Ptr{Cvoid}}, s::Status) = s.ptr
 
 function get_code(s::Status)
     code = @tfcall(:TF_GetCode, Cint, (Ptr{Cvoid},), s.ptr)
     return TF_Code(code)
 end
+
+abstract type AbstractDevice end
 
 struct DevicePart{IndexType}
     name::String
@@ -72,7 +75,11 @@ end
 device_index_from_zero(part::DevicePart{Int}) = "$(part.name):$(part.index-1)"
 device_index_from_zero(part::DevicePart) = "$(part.name):$(part.index)"
 
-struct Device
+struct NamedDevice <: AbstractDevice
+    name::String
+end
+
+struct Device <: AbstractDevice
     parts::Vector{DevicePart}
 end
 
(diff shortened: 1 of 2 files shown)


src/core.jl
```diff
@@ -72,7 +75,11 @@ end
 device_index_from_zero(part::DevicePart{Int}) = "$(part.name):$(part.index-1)"
 device_index_from_zero(part::DevicePart) = "$(part.name):$(part.index)"
 
-struct Device
+struct NamedDevice <: AbstractDevice
```
Keno: Note, I'm not sure about this particular thing. The reason I added it is that the TPU naming scheme `/job:tpu_worker/replica:0/task:0/device:TPU:0` wasn't compatible with the existing scheme (because of the last part having two colons). I don't know anything about TensforFlow device naming, so there's probably a better solution.

malmaud: It looks like TensorFlow has changed naming conventions at some point since I last looked at it and we should just modify the existing `DevicePart` to handle it, rather than adding a new device type. I'll push to this PR to do so, if you don't mind.

Keno:
Go for it.

malmaud:
I'll merge if this looks to you, @Keno .

Keno:
I'll run it in the morning to make sure it still works.

src/core.jl
```diff
@@ -150,7 +153,7 @@ struct OperationContext
     control_ops::Vector{Vector{Any}}  # Can't make Operation to break type cycle
     names::Vector{String}
     while_context::Vector{tensorflow.WhileContextDef}
-    devices::Vector{Device}
+    devices::Vector{AbstractDevice}
     is_top_level::Ref{Bool}
 end
```
oxinabox: We should fix this type, to be parametric.

src/core.jl
```diff
@@ -501,6 +509,46 @@ function Base.close(sess::Session)
...
+    end
+end
+
+struct DeviceInfo
+    name::String
+    device_type::String
+    membytes::Int64
+end
+
+Base.unsafe_convert(::Type{Ptr{Cvoid}}, l::DeviceList) = l.ptr
```
oxinabox: Can we not use single letter variable names?
In particular not using `l` which often looks like `1` or `I`.

src/core.jl
malmaud: Might as well just reduce complexity and turn that back to `Device` now that there's only device type again.

Keno:
Seems to work fine. LGTM.

PR #431, opened 2018-09-07
Update logistic.md

colbec:
Various little syntax changes to suit version 1.0; a workaround for the rounding of the loss since macro @sprintf is currently not working.

### DIFF
--- a/docs/src/logistic.md
+++ b/docs/src/logistic.md
@@ -2,13 +2,13 @@
 
 
 ```julia
-using Distributions
+using Distributions, TensorFlow, PyCall, MacroTools
 
 # Generate some synthetic data
 x = randn(100, 50)
 w = randn(50, 10)
-y_prob = exp(x*w)
-y_prob ./= sum(y_prob,2)
+y_prob = exp.(x*w)
+y_prob ./= sum(y_prob,dims=2)
 
 function draw(probs)
     y = zeros(size(probs))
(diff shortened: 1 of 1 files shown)


docs/src/logistic.md
```diff
@@ -2,13 +2,13 @@
 
 
 ```julia
-using Distributions
+using Distributions, TensorFlow, PyCall, MacroTools
```
malmaud: Why do you need PyCall and MacroTools?

docs/src/logistic.md
```diff
@@ -39,10 +39,11 @@ saver = train.Saver()
 # Run training
 run(sess, global_variables_initializer())
 checkpoint_path = mktempdir()
-@info("Checkpoint files saved in $checkpoint_path")
+@info "Checkpoint files saved in $checkpoint_path"
```
malmaud: Why get rid of the parens?

docs/src/logistic.md
```diff
@@ -39,10 +39,11 @@ saver = train.Saver()
...
 for epoch in 1:100
     cur_loss, _ = run(sess, (Loss, minimize_op), Dict(X=>x, Y_obs=>y))
-    println(@sprintf("Current loss is %.2f.", cur_loss))
+    cur_loss_rounded = round(cur_loss,digits=2)
+    println("Current loss is $cur_loss_rounded.")
```
malmaud: What's wrong with `@sprintf`? Did Julia 1.0 change its API or something?

colbec: Ditto again, it would not run, so I looked and found an old discussion about sprintf and how the community was looking for some volunteer to resolve the outstanding issues. Perhaps these are now resolved. I did a `? sprintf` on my machine and found nothing of use. So I worked around it using functionality that appeared to be available. Let me know if I can help by further digging.

docs/src/logistic.md
colbec: I just re-tested the old code and now with PyCall and MacroTools not called for the code works. I'm not able to explain why the situation was different the first time.

docs/src/logistic.md
colbec: On second test the code works with or without parens.

docs/src/logistic.md
malmaud: It's moved to `Printf` standard library module, apparently.

malmaud:
I just tweaked a few things; should be g2g now.

PR #437, opened 2018-09-18
Use MLDatasets.jl rather than MNIST.jl

smist09:
Changed the package used, so changes the dependencies as well. Note that it works for the simple linear example file, but the complicated convolutional example gives random results, not sure if it worked before or if its something else.


malmaud:
This looks good, thanks! I would like to get to the bottom of the convolution results before merging, though. 
 I'll test locally, time, allowing.

smist09:
I think actually the convolution one does work, it just takes far longer to train than I expected. I think it might be worth doing some comparison of memory usage and speed versus the same exercises in Python. I noticed in the Differential Equations youtube tutorial there was a lot of talk about being very careful of matrix operations to ensure they were element by element and didn't create a lot of intermediate matrices lying around that cause strain on the Julia heap manager.

oxinabox:
TensorFlow doesn't directly use julia's  memory manager,
during `run`.
That lives entirely in TensorFlow C land.

Only differences are the tensorflow graphs we generate maybe at little different to theirs.
I feel like fixing them to act the same is the job of XLA though.

PR #438, opened 2018-09-23
Update DockerFiles to use Julia 1.0.0

metanoid:
First attempt at getting TensorFlow.jl working on Julia 1.0.0 inside Docker containers

### DIFF
--- a/Docker/cpu/setup.jl
+++ b/Docker/cpu/setup.jl
@@ -1,5 +1,6 @@
 #Pkg.add("TensorFlow")
-Pkg.clone("https://github.com/malmaud/TensorFlow.jl.git")
+using Pkg
+Pkg.add(PackageSpec(url="https://github.com/malmaud/TensorFlow.jl.git", rev = "master"))
 Pkg.build("TensorFlow")
 Pkg.add("IJulia")
 using TensorFlow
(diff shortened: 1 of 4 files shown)


Docker/cpu/setup.jl
```diff
@@ -1,5 +1,6 @@
 #Pkg.add("TensorFlow")
-Pkg.clone("https://github.com/malmaud/TensorFlow.jl.git")
+using Pkg
+Pkg.add(PackageSpec(url="https://github.com/malmaud/TensorFlow.jl.git", rev = "master"))
```
malmaud: Now that TensorFlow.jl has stabilized, how about just `Pkg.add("TensorFlow", version="0.10")` instead of directly cloning `master`?

Docker/gpu/setup.jl
```diff
@@ -1,5 +1,6 @@
 #Pkg.add("TensorFlow")
-Pkg.clone("https://github.com/malmaud/TensorFlow.jl.git")
+using Pkg
+Pkg.clone(PackageSpec("https://github.com/malmaud/TensorFlow.jl.git", rev = "master"))
```
malmaud: Same comment here as in the CPU version.

PR #440, opened 2018-09-24
fix up version info

oxinabox:
Version info wasn't displaying right in 0.7

This should fix that


oxinabox:
This was still broken.
I think.

 I'm having trouble running locally
And was waiting for Travis to check it, but got called away.

Broken in a different way to before though.

malmaud:
I pushed a fix.

Issue #439, opened 2018-09-23
Logistic Regression example does not run in Docker container


metanoid:
I used commit 05ea5a6929ccdf28870591b56aac81795203fc23 to install Julia 1.0.0 and TensorFlow in a Docker container.  Then I tried running the "Logistic Regression" example in the resulting terminal.

* First problem:  User needs to explicitly `add`, `build`, and `using PyCall` to get rid of messages complaining of PyCall being missing
* Second Problem: `py_gradients not defined` (See error log at bottom of issue)

```
----------------
Library Versions
----------------
$(Expr(:escape, :(ENV["TF_USE_GPU"]))) = "0"
Trying to evaluate ENV["LIBTENSORFLOW"] but got error: KeyError("LIBTENSORFLOW")

$(Expr(:escape, :(tf_version(kind=:backend)))) = v"1.10.0"
Trying to evaluate tf_version(kind=:python) but got error: RemoteException(2, CapturedException(UndefVarError(:py_tf), Any[(#13 at TensorFlow.jl:181, 1), (#116 at process_messages.jl:276, 1), (run_work_thunk at process_messages.jl:56, 1), (run_work_thunk at process_messages.jl:65, 1), (#102 at task.jl:259, 1)]))
$(Expr(:escape, :(tf_version(kind=:julia)))) = v"0.10.1+"

-------------
Python Status
-------------
$(Expr(:escape, :(PyCall.conda))) = false
$(Expr(:escape, :(ENV["PYTHON"]))) = "/usr/bin/python"
$(Expr(:escape, :(PyCall.PYTHONHOME))) = "/usr:/usr"
Trying to evaluate readstring(#= /root/.julia/packages/TensorFlow/W6Dwj/src/version.jl:101 =# @cmd("pip --version")) but got error: UndefVarError(:readstring)
Trying to evaluate readstring(#= /root/.julia/packages/TensorFlow/W6Dwj/src/version.jl:102 =# @cmd("pip3 --version")) but got error: UndefVarError(:readstring)

------------
Julia Status
------------
Julia Version 1.0.0
Commit 5d4eaca0c9 (2018-08-08 20:58 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, skylake)
[...]

PR #442, opened 2018-09-24
Start Python task in current project environment.

malmaud:


malmaud:
Closes https://github.com/malmaud/TensorFlow.jl/issues/439

oxinabox:
Wait, really?
That isn't the default?
That sounds like a upstream julia bug.

malmaud:
Indeed. I was advised to try the solution in this Pr although I'm not
actually sure it works.

Issue #447, opened 2018-10-02
Writing own optimization function


askrix:
Mr. Malmaud and Mr. White,

in the actual version of "TensorFlow.jl" v0.10.2 are the two optimizers provided train.AdamOptimizer() and train.GradientDescentOptimizer(). On the other hand, TensorFlow's API provides [the list of eight additional optimizers](https://www.tensorflow.org/versions/r1.0/api_guides/python/train#Optimizers). In the end September this year emerged the publication (I will provide the link) which shows that ADAM doesn't always converge compared to SGD. For my project I'd like to implement SGD. The way to do so was discussed  on [Stackoverflow](https://stackoverflow.com/questions/38431054/how-to-create-an-optimizer-in-tensorflow). Furthermore, I found this [article](https://towardsdatascience.com/custom-optimizer-in-tensorflow-d5b41f75644a) which describes in a great detail how to implement a custom optimizer for TensorFlow in Python. My question is if it's possible to do so in Julia using "TensorFlow.jl" as well?

Many thanks in advance and best regards,
Askrix

oxinabox:
Yes, it is possible.

See the code in
https://github.com/malmaud/TensorFlow.jl/blob/master/src/train.jl

If you do implement a cool new (or old) optimize feel encouraged to make a PR.

askrix:
Mr. White,

here is [the link](https://arxiv.org/abs/1705.08292) to the article I wrote about the last week.

askrix:
Mr. White,

I chose Himmelblau's function from the list of [Test functions for optimization](https://en.wikipedia.org/wiki/Test_functions_for_optimization) and implemented _simple gradient_, _momentum_ and _NAG_ (Nesterov accelerated gradient). After testing I chose the _NAG_ and following the source code from the link you provided tried to implement the algorithm for "TensorFlow.jl".
   Regarding the implementation I have the following question. Basically NAG consist of two steps
       `ν = γ*ν + η*Δf(θ - γ*ν);`
        `θ = θ - ν;` 
i don't understand how I can enforce "grad" to evaluate with θ - γ*ν as input instead of θ. Secondly, how I can create a pull request properly? What is supposed to be "base:" and "compare:"?

oxinabox:
> i don't understand how I can enforce "grad" to evaluate with θ - γ*ν as input instead of θ.

I am not sure what the best way to achieve that is.

I don't really know much about NAG,
[...]

PR #450, opened 2018-10-12
Update train.jl

askrix:
"Nadam" and "AMSGrad" are implemented. ;-)

### DIFF
--- a/src/train.jl
+++ b/src/train.jl
@@ -183,6 +183,123 @@ function apply_gradients(optimizer::AdamOptimizer, grads_and_vars; global_step=n
     return group(ops...)
 end
 
+mutable struct NadamOptimizer <: Optimizer
+    η::Float64
+    β1::Float64
+    β2::Float64
+    ϵ::Float64
+    name::String
+end
+
+NadamOptimizer(learning_rate; β1=.9, β2=.999, ϵ=1e-8, name="nadam") = NadamOptimizer(learning_rate, β1, β2, ϵ, name)
+
+function NadamOptimizer(; η=.001, kwargs...)
+    NadamOptimizer(η; kwargs...)
+end
+
+function Base.show(io::IO, optim::NadamOptimizer)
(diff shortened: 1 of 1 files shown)


oxinabox:
This looks promising, thanks. Nice work diving into a fairly deep part of the system.
Can you add some tests?

@malmaud  should probably do the code review, I'm not very familiar with this part of the code myself.

askrix:
Mr. White,

many thanks for your feedback. I tried my best to get the implementation done. As I wrote you about  it in [issue #447 ](https://github.com/malmaud/TensorFlow.jl/issues/447#issuecomment-428957209), I tested the implementation on my computer using Himmelsbau's function and compared the number of iterations each algorithm needed to converge by the given precision and other parameters. 

How should look a test for the algorithms in "TensorFlow.jl"? Should I write for example a linear regression task, where different optimization algorithms are benchmarked against each other?

In connection with my pool request I'd like to ask if it will be possible to reorganize the source code and to export all of the optimization algorithms from "train.jl" in the future. For example by creating an extra sub-folder so that each optimizer will have an own file. (E.g. "Nadam.jl"). I'm asking for it because I'm planning to add `doc""" some explanations and notes """` to reference to the publications on which my implementations are based. 

Best regards,
Askrix

askrix:
@malmaud Mr. Malmaud,

so far I still didn't get any further feedback regarding my pull-request. I guess my implementation is either not sufficient good or/and something is missing. Could you tell me what I should correct or/and add to the code?

In the meanwhile I prepared the following optimizers: AdaGrad, Adadelta, RMSProp and AdaMax. Should I open a new pull request?

Best regards,
Askrix

malmaud:
Sorry for the delay, @askrix. These actually look really good! We need some kind of test, even if it's informal, to make sure these work. 

I propose adding tests that run the 'logistic.jl' example, but with your new optimizers instead of the Adam optimizer used now. Then we can at least check that the error goes to near zero. You could just modify `logistic.jl` to be a function that takes an optimizer constructor as an argument and then modify `runtests.jl` to pass in all the new optimizers to the logistic example. I'm open to other testing ideas as well.

How does that sound?

PR #458, opened 2018-10-28
Add equality overrides for `Device` and `DevicePart`

staticfloat:
This is needed for our XLA work, so that we can ensure that the device a
tensor is coming fron matches the device a model is running on.


malmaud:
Thanks! Would it be feasible to use AutoHashEquals.jl here instead of the manual approach?

staticfloat:
Looks like it, thanks for the pointer.

malmaud:
Great.

PR #463, opened 2018-11-04
Fix types in set_attr_shape_list

Keno:
The lack of `Cint` for the second to last argument is particularly problematic, because TensorFlow would get incorrect values for the lengths of the shape array.


malmaud:
Great.

PR #467, opened 2018-11-11
Make unary "-" work with @tf

MoonCoral:
As is, it errors because of the `name` keyword when used with `@tf`.

### DIFF
--- a/src/ops/math.jl
+++ b/src/ops/math.jl
@@ -167,7 +167,9 @@ function Base.round(::Type{T}, value::AbstractTensor) where T
 end
 
 
--(n::AbstractTensor; kwargs...) = negative(n; kwargs...)
+function -(n::AbstractTensor; kwargs...)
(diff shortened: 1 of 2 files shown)


oxinabox:
Thanks for this. I didn't know it was borked.

Can you add a test for this in test/meta.jl

Also I am pretty sure it needs a `@op` before the `function`.

MoonCoral:
Not sure why Travis is segfaulting in the python 3 instances, I'm using python 3 locally and everything related to this PR seems to work.

oxinabox:
I have kicked it to retry.
It looks right to me.

MoonCoral:
The same failure seems to happen since a month. See https://github.com/malmaud/TensorFlow.jl/pull/435

oxinabox:
Well fair enough,
I've openned another PR to track that  #468

PR #468, opened 2018-11-12
Allow failure on Python 2, not 3

oxinabox:
Since Python 3 is now the default
https://github.com/JuliaPy/Conda.jl/pull/108


oxinabox:
This basically undoes #435

oxinabox:
Indeed, but this can track that.

malmaud:
@oxinabox trying to clean up the PRs. Do we still want this? I kinda forget the context.

oxinabox:
We do still want this

PR #478, opened 2018-12-17
Incorporating Optim.jl first order solvers

kailaix:
I have wrote a wrapper for optimizers in Optim.jl. Although the implementation is not optimal in efficiency and the APIs cannot be made consistent with the current one, it is nice to have optimizers such as L-BFGS, BFGS, Accelerated Gradient Descent and Conjugate Gradient Descent. Some other minor issues are fixed. Test cases passed on my mac.


oxinabox:
This is one of the most awesome things I have seen, great idea.

I am not sure if it belongs in TensorFlow.jl itself,
or if it would be better to make a seperate package (that depends on TensorFlow and Optim),
to hold it.

What do you think @malmaud ?

src/ops/transformations.jl
```diff
@@ -380,16 +380,16 @@ Returns:
     local result
```
oxinabox: The changes to this file should be a seperate PR. :-)

oxinabox:
Are you familar with how to use `git cherrypick` to split a branch (PR) into multiple branches?

malmaud:
This is really cool, thanks! @oxinabox is right that the parts of the PR that do the Optim integration should be separate from the generic improvement parts. I think we can put it in TensorFlow.jl for now, then later when we refactor into TensorFlowCore.jl and the more feature-rich package, this functionality will end up in the latter.

kailaix:
Hi, I have seperated the PR into multiple PRs.  This PR can be closed.

PR #480, opened 2018-12-17
Convolution in 1D

kailaix:
Convolution in 1D. The implementation follows https://github.com/tensorflow/tensorflow/blob/r1.12/tensorflow/python/ops/nn_ops.py


malmaud:
Thanks, this is great! Sorry for the long delay. I've created a rebased version of the PR in https://github.com/malmaud/TensorFlow.jl/pull/506.

PR #493, opened 2019-01-29
Make it possible to run from Enviroments that do not directly have Pycall

oxinabox:
This adds a Project.toml
so that we can run the workers in the TensorFlow enviroment,
which will have PyCall for sure.

When making the Project.toml i did not both to transcribe the Compat rules,
since they are basically not julia 0.7 compat anyway.


malmaud:
Cool cool. Is there a guide somewhere on how Project.toml and the whole environment-based package system works?

oxinabox:
Pkg.jl docs.

It could be better, but I have tried and I can't make it better yet.

malmaud:
OK, I get what this PR does now. Looks good, merge when you're ready.

PR #494, opened 2019-01-30
Stop segfaulting when displaying a bunch of controlflow related shapes

oxinabox:
Closes #466

That was a fiddly one.

PR #495, opened 2019-01-30
Do not break showing the shape of Variables

oxinabox:
Follow up on #494

### DIFF
--- a/src/core.jl
+++ b/src/core.jl
@@ -1793,7 +1793,7 @@ function set_tensor_shape(tensor::AbstractTensor, dims)
 end
 
 function get_tensor_num_dims(tensor::AbstractTensor)
-    if tensor.op.op_name == "NoOp"
+    if tensor isa Tensor && tensor.op.op_name == "NoOp"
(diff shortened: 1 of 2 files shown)


src/core.jl
```diff
@@ -1793,7 +1793,7 @@ function set_tensor_shape(tensor::AbstractTensor, dims)
 end
 
 function get_tensor_num_dims(tensor::AbstractTensor)
-    if tensor.op.op_name == "NoOp"
+    if tensor isa Tensor && tensor.op.op_name == "NoOp"
```
malmaud: How about just adding a new method to `get_tensor_num_dims` instead of the `isa` check?

oxinabox: Could do, but other than this bit it would be the same.

The only reason to do the `isa` is that `Variable`s do no have a `.op` field so we can't check if they are `"NoOP"`.
This whole condition will compile away during specialisation.

malmaud:
Ok, I'm sold.

oxinabox:
This is now ready to merge

PR #505, opened 2019-05-08
Adding Windows enabled TensorFlow

blair0011:
The CPU version passes all but the TensorBoard test, but it has path issues which I really don't care about. If I want to use TensorBoard I will run it in Python. The GPU passes 90% of the tests but there are a couple that are throwing errors  because of badly formatted input data. The Tensors are not being created in the shape that the model expects. Not sure how to fix this, need more time to investigate but for now most usages do work. Just don't use the logistic example with the GPU enabled system it will core dump.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "TensorFlow"
 uuid = "28f6a940-23b8-11e9-3177-cd2b159d7b44"
 authors = ["Jon Malmaud (@malmaud)", "Lyndon White (@oxinabox)"]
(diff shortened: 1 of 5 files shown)


oxinabox:
@blair0011  was this closed by mistake?

blair0011:
I updated my Julia system and broke everything around tensorflow so want to make sure that doesn't happen to anyone else. I need to figure out why I am getting crashes even for the CPU version.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #506, opened 2019-05-17
conv1d

malmaud:
Rebased version of https://github.com/malmaud/TensorFlow.jl/pull/480


REQUIRE
```diff
@@ -13,3 +13,4 @@ MacroTools 0.3.6
 AutoHashEquals 0.1.0
 MLDatasets 0.3.0
 SpecialFunctions 0.7.0
+Optim 0.17.0
```
oxinabox: We can now delete the REQUIRE file

src/TensorFlow.jl
```diff
@@ -135,6 +135,7 @@ with_tape
 
 
 using Distributed
+using Optim
```
oxinabox: Why is this in This PR?

PR #523, opened 2021-04-01
Typos and version used

PallHaraldsson:
[skip ci]

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,8 @@ A wrapper around [TensorFlow](https://www.tensorflow.org/), a popular open sourc
 
 | ⚠️ Notice ⚠️ |
 | --- |
-| **Tensorflow.jl is in minimal maintance mode** |
-| While it works, it is not receiving new features, and is  bound to an old version of libtensorflow. |
+| **Tensorflow.jl is in minimal maintenance mode** |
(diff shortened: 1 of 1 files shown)


oxinabox (approved):

PallHaraldsson:
That was quick, but there's another one, "maintenance" and more serious issue.

PallHaraldsson:
While you're here, should this package state something stronger than "TensorFlow.jl is in minimal maintenance mode", it's very minimal as in zero, and 1.15.5 has security issue and the issue before. I'm not sure it applies to this package, even if the older versions are also affected.

PallHaraldsson:
Say something like: "TensorFlow.jl uses 1.13.x and no update to the wrapper is planned, while current official versions are 2.x and 1.15.x"?

I'm not sure if something like that is enough.

https://github.com/tensorflow/tensorflow/releases/tag/v2.4.1

Tensorflow.jl is in minimal maintance mode"

PallHaraldsson:
Maybe just merge this now as is?

PallHaraldsson:
Just merge this? Was approved before I added the version info.

malmaud:
Sure thanks, sorry for the delay.

PR #526, opened 2024-04-30
Recommend Lux.jl

PallHaraldsson:
It seems better (shares lots of code with Flux). Maybe rather point to https://github.com/LuxDL/Lux.jl (and Flux? or not, to not confuse people too much here...?)


oxinabox:
I disagree.
Flux is better for general ML user.
Lux is better for SciML type user.
Both are *ok* for both.


But I don't think we should get into the habbit of updating the readme of a deprecated package for ever with the latest recommendations.
Folk can see the the last edit date.

